PR #8, opened 2015-02-26
added pa update

cAP5L0CK:


cAP5L0CK:
also changed "level" to "amount" where appropriate.  pa was missing before because of this.

PR #18, opened 2015-03-12
New section and, website, grammar question

davidlatapie:
Added Goods after discussion with fluffypony
Added cryptonit.net under this section
Used the esperanto form of monero. This has implication, because so far we avoided the matter of choosing because we avoided the plural. This doesn't mean only English plural or only Esperanto plural is accepted - both could, like both color and colours are accepted. Still, style-wise, it is better to stick to one word (color everywhere or colour everywhere) so this question still remains for -s vs -j.

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -30,3 +30,7 @@
       url: http://www.cryptoescrow.eu
     - name: MoneroClub OTC Monero Trading
       url: https://www.moneroclub.com/offers/
+- category: Goods
+  merchants:
+    - name: Physical bitcoins and moneroj
+      url: https://cryptonit.net
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -32,5 +32,5 @@
       url: https://www.moneroclub.com/offers/
 - category: Goods
   merchants:
-    - name: Physical bitcoins and moneroj
+    - name: Physical Monero and Bitcoin coins
       url: https://cryptonit.net


fluffypony:
Why not use it as an adjective for this?

"Physical Monero coins"

davidlatapie:
Yep, forgot there is no plural for adjective in English. This only postpone the matter though (not that I have problem with postponing - urgent first, less urgent later).
We shall mention Bitcoin coins too, we should not filter the content that a website sell (as long as it is legal, of course)

"Physical Monero and Bitcoin coins"

fluffypony:
My personal opinion is that a plural can be skipped completely:

- "That'll be 5 Monero, sir"
- "How many Monero do I owe you?"
- "They sell Monero on Poloniex"
- "The vast quantity of Monero available"

I can't (yet) think of a sentence where plurality is necessary.

davidlatapie:
Don't we say "That'll be 5 dollars, sir"?

/me thought he understands English plural

fluffypony:
Yes we do, but not every currency name behaves like that. Take French Francs, for example:-P

A better example is the Yen. We don't say "that'll be 5 Yens, sir", we just say "that'll be 5 Yen, sir", and then karate-chop them into oblivion.

davidlatapie:
OK so let's go with Physical Monero and Bitcoin coins (ugly "coin-coin" here, but Monero has to be first!)

fluffypony:
Sounds good - if you make the changes to your patch-5 branch and push them then this PR will update and I can merge it

davidlatapie:
Seems to be done. I'll wait confirmation before closing, though.

fluffypony:
Perfect:)

PR #20, opened 2015-03-17
GPG informations

davidlatapie:


fluffypony:
I think we should put your GPG key in the source code instead, as there's a GPG trust chain for all commits / merges there? You'll still be able to hotlink it, e.g.: https://raw.githubusercontent.com/monero-project/bitmonero/master/utils/gpg_keys/fluffypony.asc

I agree on having the GPG keys / fingerprints attached to the people page, but it's starting to get a bit messy now - let me see if I can clean up the formatting a bit.

davidlatapie:
What about "[[GPG](https://raw.githubusercontent.com/monero-project/bitmonero/master/utils/gpg_keys/fluffypony.asc)]"?

PR #26, opened 2015-03-28
field definitions are a thing

Gingeropolous:
yay


fluffypony:
H4 tags should look like ```#### Sub-header Goes Here```, and titles should be appropriately capitalised:)

PR #48, opened 2015-07-24
Flesh out stealth addresses a bit

luigi1111:
Sometimes trying to break stuff like this down results in awkward sentences. Feedback pls.

### DIFF
--- a/knowledge-base/moneropedia/stealthaddress.md
+++ b/knowledge-base/moneropedia/stealthaddress.md
@@ -2,9 +2,11 @@
 layout: moneropedia
 entry: "Stealth Address"
 terms: ["stealth-address", "stealth-addresses"]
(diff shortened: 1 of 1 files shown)


fluffypony:
The summary text shows up on hover, so it has to be extremely basic and has to assume the person knows nothing else about Monero. I don't think they'd necessarily grasp the concept of unlinkability or how randomness is good just from reading the words.

Also, instead of hot-linking to the Moneropedia article, just use an @ mention, as in \@ring-signatures or \@blockchain

luigi1111:
I'm a noob. Will force push later.

luigi1111:
Ok, I changed it. I'm not sure exactly how "@transaction" will show up. I went with "automatic" instead of unlinkable/random.

PR #53, opened 2015-08-09
added android wallet to homepage

dternyak:


fluffypony:
I think this should be added to the Choose A Monero Client page instead (https://getmonero.org/getting-started/choose) - that front page is going to change to only recommending the core client as soon as we have finished the GUI, so anything else is just a temporary usability hack;)

PR #102, opened 2016-02-13
deletion of cryptsy from list of merchants

stefioan:


Hueristic:
But they might re-open! LOL

At least anyone that reads our threads knew enough not to use that scam exchange, :)

PR #129, opened 2016-07-20
Duplicate entry on merchants.

stefioan:
Duplicate entry of "Web Developer - Stefanos".


coolcodecrypto:
This PR appears to be deleting the incorrect duplicate entry (assuming the alphabetical order is meant to be retained).

fluffypony:
@coolcodecrypto we can sort out ordering in the code, doesn't have to be ordered in-list - thanks for the heads up

PR #143, opened 2016-09-02
Update restore_account.md

palexande:
Change formatting of command line argument.


palexande:
I'll rewrite this so it makes sense with the new binaries and tidy it up a bit.

PR #156, opened 2016-09-17
Update blockchain.md

EWBears:
Fleshed on definition


fluffypony:
Good job, but it needs to be moved down - look at the other Moneropedia entries. The summary is EXTREMELY short, because it's a summary that shows when you hover your mouse over the word. You can flesh it out properly below the YAML header:)

EWBears:
Ok thanks for the feedback.  I didn't know about the definition showing on mouse over.  

I will go ahead and move the fleshed out version in the proper place and then start expanding on the other entries!

EWBears:
OK I think that I corrected the previous errors.  Please let me know if anything is still wrong!

PR #160, opened 2016-09-20
Update restore_account.md

palexande:
Changed formatting, added steps to support new functionality, replaced simplewallet with monero-wallet-cli.


kannix:
the headline says it is for all platforms but you changed some lines to include the .exe suffix
in my opinion it would be better to omit that

palexande:
Yeah, you are right, I will remove that.

PR #164, opened 2016-09-22
Create ringCT.md

palexande:


ChristopherKing42:
Perhaps a link to [MRL-0005](https://lab.getmonero.org/pubs/MRL-0005.pdf) would be good?

palexande:
Alright.  I will probably put the abstract and then the link under an advanced heading.

PR #211, opened 2017-01-08
Removed dead links and Updated Creative Commons license

alvinjoelsantos:


alvinjoelsantos:
[CC BY-SA 4.0](https://creativecommons.org/licenses/by-sa/4.0/) is the same license used by Wikipedia and would require any derivative work to be offered under the same, similar or a compatible license. However, if the devs do not want to add such restriction, [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/) could be used instead.

PR #236, opened 2017-03-11
Update view_only.md

mjsjml:
The method for making the view only wallet in Ubuntu is now simpler than before.  The only argument after --generate-from-view-key is now the wallet name.  The program will prompt for the other fields.


amiuhle:
Related / duplicate: #235

mjsjml:
Haha. I'm new at this type of thing.

fluffypony:
Closed in favour of 235

PR #243, opened 2017-03-28
Spanish translation

romanturbo:
Spanish Translation. Everything should be right. Im planning in translating everything to spanish.


fluffypony:
You can't just edit the text in the tree to create a translation, unfortunately - with the redesign we're sorting out a translation setup that will led you edit the files directly.

PR #252, opened 2017-04-18
Update python package - pymonero to MoneroPy

bigreddmachine:
I've deprecated `pymonero` and have began merging the functionality of it into `MoneroPy`, which is actively maintained and undergoing development. `pymonero` therefore should not be listed anymore.


SamsungGalaxyPlayer:
@bigreddmachine Can I suggest you update the pymonero GitHub readme to say that it is no longer updated and include the link for MoneroPy?

bigreddmachine:
Yeah, good idea.

PR #253, opened 2017-04-18
rm MoneroWorld explorer and add two others

bigreddmachine:
MoneroWorld no longer hosts an explorer, but two other are now maintained based off the same onion explorer source.


SamsungGalaxyPlayer:
@fluffypony I think this is ready to be merged.

PR #254, opened 2017-04-19
Add Unofficial Linkability Response

SamsungGalaxyPlayer:
This time, the images and formatting is correct to my knowledge (thanks ajs!).

Some of the commits, listed below, involve adding and deleting the same file. I do not know how to remove these from the pull request. My apologies:

8892a73
7033bb7
18bb205
789c9e9

Thus, the two commits to worry about are:

05fdd59
1634b0d


SamsungGalaxyPlayer:
@iamsmooth can you please look over this now? I made some changes that I would like you to verify are accurate.

SamsungGalaxyPlayer:
@Gingeropolous is this better? It's hard for me to add what you explained to a specific section.

iamsmooth:
@Gingeropolous the pool payouts are traceable to the pools coinbases. This is often true whether or not the pools used positive mixins or not (though at least one pool had a more complex payout scheme that involves moving coins to an intermediate wallet first, and would reduce this traceability to some extent).

Tracing (or non-tracing) via mixins refers to _preceding_ transactions. In the case of pool payouts that is the coinbase.

There is another question whether users _spending_ those pool payouts would also use mix 0, and in most cases, that is very likely to be the case as well, but that is more of a user choice than anything else.

SamsungGalaxyPlayer:
My apologies for the poor GitHub use. I think the actual paper is ready for release now. Can you please review it, @iamsmooth @Gingeropolous @fluffypony 

Don't merge this in the next few hours. I'd like to give Andrew Miller a chance to point out any other mistakes, which is something I promised to him.

fluffypony:
Been more than a few hours, merging it

Issue #256, opened 2017-04-25
Please add a new site to the Monero Merchants and Services Directory


AzurSamui:
Azur Samui, a luxury apartment and villa development on Koh, Samui Thailand now accepts Monero in full for purchases. Can you please add to the directory? (AzurSamui.com) Thanks!

alvinjoelsantos:
Link to website added, please close issue.

AzurSamui:
Thanks!

PR #257, opened 2017-04-25
added new merchant (Azur Samui)

alvinjoelsantos:
Issue: #256


SamsungGalaxyPlayer:
I'm conflicted. Should this go under a good? You could use Monero to buy a tangible thing (house/apartment).

alvinjoelsantos:
Initially, I was thinking about making a new category for Real Estate. However, it is my understanding that foreigners can't own property outright in Thailand and what they offer are shares in a managment company and handle all the paperwork, which is more like a service.

PR #263, opened 2017-05-08
Logs for the Kovri and Dev meetings held on 2017-05-07

dEBRUYNE-1:


dEBRUYNE-1:
Added "Logs for the Kovri meeting held on 2017-04-23".

Issue #268, opened 2017-05-23
New Merchant: cloakVPN.com


pwrcycle:
Please add cloakVPN.com to the Monero Merchants and Services Directory
Users are automatically provisioned based on the Payment_ID sent in the Monero transaction.

[http://cloakVPN.com/monero](http://cloakVPN.com/monero)

Twitter announcement
[https://twitter.com/cloakVPN/status/861638668909334528](https://twitter.com/cloakVPN/status/861638668909334528)
Reddit announcement
[https://www.reddit.com/r/Monero/comments/6cpfd2/cloakvpncom_now_accepts_monero_using_payment_id/](https://www.reddit.com/r/Monero/comments/6cpfd2/cloakvpncom_now_accepts_monero_using_payment_id/)

contact admin@cloakVPN.com

rehrar:
@pwrcycle can we close this issue? :) It's done.

pwrcycle:
Resolving :)

Issue #239, opened 2017-03-27
 Please add a new site to the MONERO MERCHANTS AND SERVICES DIRECTORY 


elivantaylor:
Hello,

Our company Ivan Multimedia Productions now accepts MONERO as payment fro our music
products.

We are a independent music publishing company that sell our own music worldwide online.
Our web site address is http://ivanmultimediaproductions.webs.com
Buyers get $3.49 off  each order if they order using Cyptocurrency.

Thanks,

el-Ivan Taylor
Ivan Multimedia Productions
http://ivanmultimediaproductions.webs.com

PR #270, opened 2017-05-28
Guidelines for merchants

serhack:
In the new website, I will write some guidelines for merchants. As now, I wrote this line. 
This pr add new merchants websites.
Issues fixed:
- #268
- #239 

Cheers

### DIFF
--- a/getting-started/merchants.md
+++ b/getting-started/merchants.md
@@ -35,4 +35,4 @@ attribution: "<!-- Icon is based on work by Freepik (http://www.freepik.com) and
 
 </div>
 
-All of the merchants are organised alphabetically. If there are any merchants that no longer provide Monero services, or a merchant that wishes to be added, please [open an issue on this website's Github repository](https://github.com/monero-project/monero-site/issues) to alert us.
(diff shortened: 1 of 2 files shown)


getting-started/merchants.md
```diff
@@ -35,4 +35,4 @@ attribution: "<!-- Icon is based on work by Freepik (http://www.freepik.com) and
 
 </div>
 
-All of the merchants are organised alphabetically. If there are any merchants that no longer provide Monero services, or a merchant that wishes to be added, please [open an issue on this website's Github repository](https://github.com/monero-project/monero-site/issues) to alert us.
+All of the merchants are organised alphabetically. If there are any merchants that no longer provide Monero services, or a merchant that wishes to be added, please [open an issue on this website's Github repository](https://github.com/monero-project/monero-site/issues) to alert us. Please, We can't accept illegal, not-trusted or scammed e-shop.
```
jonathancross: I don't think this last line should be added as it has multiple issues.

1. "W" in "We" should not be capitalized.
2. Beginning with "Please," is awkward.
3. We may not want to support illegal shops, however is that really a problem here?  Remember what is illegal in one place could be legal somewhere else as well.
4. "not-trusted" is vague -- by who?
5. "scammed e-shop" means a **shop** that was scammed by somebody else.

serhack: Hello, with new changes of website, I can close the pull :D

PR #293, opened 2017-07-04
Reverted FFS links

rehrar:
- Reverted link in Contributing to forum.getmonero.org
- Reverted header links to forum.getmonero.org


jonathancross:
@snaggen Now fixed in #305

PR #309, opened 2017-07-09
First revision of pool mining user guide

mattcode55:
* Add guide to user-guides index
* Create first revision of guide
* Add four screenshots to make guide easier to read


resources/user-guides/mine-to-pool.md
```diff
@@ -0,0 +1,114 @@
+---
+layout: static_page
+title: How to mine on a pool
+---
+# Selecting a pool
+
+There are many pools to choose from, a list is available at
+[moneropools.com](https://moneropools.com). Mining on a larger pool could mean
```
mattcode55: >Do not recommend certain mining pools, miners, etc. Instead, use placeholders. Perhaps give an example with pool.getmonero.org as a fictitious mining pool.

I hope mentioning moneropools.com is okay. I decided that it would be better than maintaining a list of pools on this site, and it would prevent people going onto dodgy sites they find on Google.

resources/user-guides/mine-to-pool.md
```diff
@@ -0,0 +1,114 @@
...
+more frequent payouts, but mining on a smaller pool helps to keep the network
+decentralised.
+
+# Selecting a CPU miner
+
+Just like pools, there are a lot of miners to choose from. The one that you
+should pick depends on the hardware you want to mine on. This guide will only
+use a CPU miner, and will be using
+[xmr-stak-cpu](https://github.com/fireice-uk/xmr-stak-cpu).
```
mattcode55: Again, I hope that mentioning and linking to this miner is okay.

resources/user-guides/index.md
```diff
@@ -67,7 +67,7 @@ title: "User Guides"
                 </div>
 <div class="row start-xs text-center" markdown="1">
 
-User Guides for Mining coming soon.
+[How to mine on a pool](mine-to-pool.html)
```
erciccione: I would change it in ```How to mine on a pool (with xmr-stak-cpu)```, because the guide is very specific and cannot be used for another miner

resources/user-guides/mine-to-pool.md
erciccione: It shouldn't be a problem, but might be a good idea to nominate a couple of alternatives with links, like Cpuminer [tpruvot](https://github.com/tpruvot/cpuminer-multi/tree/master) and [Wolf](https://github.com/OhGodAPet/cpuminer-multi)

resources/user-guides/index.md
mattcode55: Yeah I agree it should be changed. The procedure is almost the same for
the xmr-stak GPU miners, so should I just change the title to "How to
mine on a pool with xmr-stak-*" or similar?

resources/user-guides/mine-to-pool.md
erciccione: shouldn't be a problem, developers often redirect to moneropools when somebody ask to suggest a pool

mattcode55:
Yeah I agree it should be changed. The procedure is almost the same for
the xmr-stak GPU miners, so should I just change the title to "How to
mine on a pool with xmr-stak-*" or similar?

resources/user-guides/index.md
erciccione: Maybe something like 'how to mine on a pool with xmr-stak (cpu or gpu)' ? . Or you can just add a quick reference to stak-gpu in the body

PR #315, opened 2017-07-10
Fix 'Open Sans' @font-face declaration

jrob-io:
The @font-face declaration has an erroneous `sans-serif` which prevents the font from being displayed (if you don't have the font installed locally).

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -29,7 +29,7 @@
 }
 
 @font-face {
-  font-family: 'Open Sans', sans-serif;
+  font-family: 'Open Sans';
   font-weight: 400;
   font-style: normal;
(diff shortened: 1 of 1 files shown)


rehrar:
I don't understand, common practice is to put the specific font that you want first under the `font-family` attribute, with more generic font families afterwards in case the font cannot be loaded. In this case `sans-serif` is identifying the generic font family to load in the event that Open Sans cannot be loaded.

jrob-io:
You're correct, but only when you go to use it on an element. When you're inside of `@font-face`, the `font-family` attribute is used to reference the font.

```
@font-face {
  font-family: "FontReference";
  ...
}

h1 { 
  font-family: "FontReference", sans-serif;
}
```

PR #329, opened 2017-07-18
Update roadmap.yml

SamsungGalaxyPlayer:
Added a few things and switched the day before the month.

### DIFF
--- a/_data/roadmap.yml
+++ b/_data/roadmap.yml
@@ -1,60 +1,60 @@
 - year: 2014
   accomplishments:
     - name: Launched on Bitcointalk
-      date: April 18, 2014
+      date: 18 April 2014
       status: completed
     - name: Renamed from Bitmonero to Monero
-      date: April 23, 2014
+      date: 23 April 2014
       status: completed
     - name: Recovered from a spam attack
-      date: September 4, 2014
+      date: 4 September 2014
       status: completed
     - name: Monero Research Lab Papers 1 and 2 published
-      date: September 12, 2014
+      date: 12 September 2014
       status: completed
     - name: Monero Research Lab Paper 3 published
-      date: September 25, 2014
+      date: 25 September 2014
       status: completed
     - name: 0.8.8.6 released
-      date: December 8, 2014
+      date: 8 December 2014
       status: completed
 - year: 2015
   accomplishments:
     - name: Monero Research Lab Paper 4 published.
-      date: January 26, 2015
+      date: 26 January 2015
       status: completed
 - year: 2016
   accomplishments:
     - name: 0.9.0 Hydrogen Helix published
-      date: January 1, 2016
+      date: 1 January 2016
       status: completed
     - name: Monero Research Lab Paper 5 published
-      date: February, 2016
+      date: February 2016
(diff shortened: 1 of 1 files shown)


_data/roadmap.yml
```diff
@@ -1,60 +1,60 @@
 - year: 2014
   accomplishments:
     - name: Launched on Bitcointalk
-      date: April 18, 2014
+      date: 18 April 2014
```
anonimal: >April 18, 2014
>18 April 2014

Converting from American English to British is a trivial moot point. I see no reason to change something that can be so easily converted back to its original format while keeping functionality intact. Why make these changes?

_data/roadmap.yml
```diff
@@ -63,17 +63,29 @@
...
       date:
       status: ongoing
-    - name: Kovri Alpha release
+    - GUI port to android
       date:
       status: ongoing
-    - name: Website Redesign
+    - name: Kovri alpha release
+      date:
+      status: ongoing
```
anonimal: s/ongoing/upcoming/

_data/roadmap.yml
QuickBASIC: >  I see no reason to change something that can be so easily converted back to its original format while keeping functionality intact.

Consistency. All the other dates are in that format currently.

anonimal: Care to point out where? A simple grep of monero-site from 2014-2017 shows that this is not the case. In fact, the majority of dates are labeled YYYY-MM-DD.

QuickBASIC: You're right. My bad. I was on mobile and thought that I had seen the dates formatted that way previously.

SamsungGalaxyPlayer: We should probably switch to a more international standard, even if it means updating the other pages. Most of the world uses DMY https://en.wikipedia.org/wiki/Date_format_by_country

_data/roadmap.yml
SamsungGalaxyPlayer: @anonimal can I change the value to `s/ongoing/upcoming/` and it will look ok on the page?

_data/roadmap.yml
QuickBASIC: I would make a case for YYYY-MM-DD even though it's second most common in the world by population... It's also used as secondary format in many countries next to the most popular and it leaves no room for confusion.

fluffypony: I second that - YMD leaves no room for confusion, DMY will be interpreted as MDY by Americans.

fluffypony:
@SamsungGalaxyPlayer plz see comments above, especially on the date format

_data/roadmap.yml
rehrar: he means 'substitute ongoing for upcoming'. That's what that format thing means. I had to learn that too.

fluffypony:
^^^ still waiting on this

fluffypony:
@SamsungGalaxyPlayer don't forget about this plx

SamsungGalaxyPlayer:
@fluffypony I finally got around to this. Sorry for the delay.

PR #334, opened 2017-07-20
Add user guide for using a remote node in the GUI wallet

cryptochangements34:


cryptochangements34:
A fix in the image path needs to be made, i will close and re-submit a similar PR

PR #335, opened 2017-07-20
Add user guide for using a remote node

cryptochangements34:
This PR adds a user guide for finding an open remote node with moneroworld.com, then connecting to that node within the GUI wallet. Also includes 1 image.


cryptochangements34:
Will do!

PR #344, opened 2017-07-28
Use absolute units with background-position

### LINKED ISSUE #324: Icon glitch in "Downloads" section
There is a reproducible icon glitch in the Downloads section, when using a larger or smaller than default font size in the browser settings. Tested on current Chrome and Firefox.

2017-07-16 20_52_58-monero - secure private untraceable

jrob-io:
Fixes #324.

`1rem` will normally convert to `16px`, but it's a relative unit that is based on the font size of `<html>`. If the user's browser overrides the default font size, it'll change the value of `1rem` and affect how the icons are taken from the spritesheet. Switching to `px` when using `background-position` will prevent this.


rehrar:
A good suggestion. Thank you so much.

PR #355, opened 2017-08-22
Make setup easier and more repeatable by using Bundler

qertoip:
Bundler makes installing gems easier and more repeatable between programmers by specifying the precise versions in a Gemfile.lock. Bundler is the expected way to setup any kind of Ruby projects and tools.


qertoip:
This pull request accidentally brings unrelated commits. Will close and resubmit.

PR #358, opened 2017-08-22
Fix broken link to the guide on downloads page

qertoip:


mattcode55:
I think that it should link to https://getmonero.org/get-started/using/ instead.

qertoip:
@mattcode55 I do agree that ideally it should link to an internal page on the getmonero.org.

However, in my opinion the suggested page does not attempt to answer "which wallet application" to use. I could not find a better page.

Meanwhile, the suggested SE answer seems to be the best resource on the topic the community has to offer.

mattcode55:
>I could not find a better page.

Maybe it would be nice to create a new page on the website that expands on the answer given on Reddit.

qertoip:
@mattcode55 Unfortunately I don't feel competent enough just yet to create a resource on the topic. Could we accept this PR simply to fix the broken link for now?

QuickBASIC:
I agree with @mattcode55 the link should be internal to the site. If not maybe we should just remind the first sentence about the guide and as it later when there's more options. Right now it's GUI, CLI, MyMonero, and someone is working on an Android wallet that's not ready to be used outside the testnet yet, so there's not much choice anyway.

PR #314, opened 2017-07-10
Update _strings_en.yml

QuickBASIC:
I feel like the language used on the main site is either wordy, confusing, or is written in a more casual writing style. We need to make sure that the information that's eventually translated is good technical writing (simple and easy to understand). It can be off-putting for new users for the language on the site to be less professional.

PR #359, opened 2017-08-22
Update Home Page

tyrionmcmaster:
Please review and comment.

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,8 +20,8 @@ Title: Home
                         <div class="main-info">
                             <h1 id="main-h1">Monero</h1>
                             <h2 id="main-h2">Private Digital Currency</h2>
-                            <p id="main-text">Monero is a secure, private, untraceable currency. It is open-source and freely available to all. With Monero, you are your own bank. Only 
-                            you control and are responsible for your funds; your accounts 
+                            <p id="main-text">Monero is a secure, private and untraceable cryptocurrency. It is open-source and accessible to all. With Monero, you are your own bank. Only 
+                            you control and are responsible for your funds. Your accounts 
                             and transactions are kept private from prying eyes.</p>
                             <p><a href="/get-started/using/" class="btn-link btn-auto btn-primary">Get Started</a></p>
                         </div>
@@ -40,7 +40,7 @@ Title: Home
                     <div class="row middle-xs info-block-row private">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12  why-text">
                             <h3>Monero is secure.</h3>
-                            <p>Monero can't be hacked to steal your funds, due to the power of distributed consensus, which you can read about here. This means that you are responsible for your money, and don't have to trust any entity to keep it safe for you.</p>
+                            <p>Monero cannot be hacked to steal your funds, due to its decentralized nature. Only you are responsible for securing your Monero.</p>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-4 col-xs-12 last-sm first-xs center-xs">
                             <img src="/img/secure-monero.png" alt="Secure safe" class="main-icon">
@@ -52,14 +52,14 @@ Title: Home
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is private.</h3>
-                            <p>The power of the blockchain usually increases security at the cost of privacy, but with Monero's sophisticated privacy-centric technology, you get all of the security benefits of the blockchain without any of the privacy trade-offs.</p>
+                            <p>Security of the blockchain normally increases at the cost of privacy. Monero is based on the Cryptonote application layer protocol, which allows for the security benefits of blockchain technology without any of the privacy trade-offs</p>
                         </div>
                         
                     </div>
                     <div class="row middle-xs info-block-row">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is untraceable.</h3>
-                            <p>By taking advantage of ring signatures, Monero makes it ambiguous which funds have been spent, and thus extremely unlikely that a transaction could be linked to particular user.</p>
+                            <p>Sending and receiving addresses as well as transacted amounts are obfuscated. Transactions cannot be linked to a particular user or real-world identity.</p>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-4 col-xs-12 last-sm first-xs center-xs">
                             <img src="/img/untraceable-monero.png" alt="Untraceable Monero paths" class="main-icon">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -20,8 +20,8 @@
                         <div class="main-info">
                             <h1 id="main-h1">Monero</h1>
                             <h2 id="main-h2">Private Digital Currency</h2>
-                            <p id="main-text">Monero is a secure, private, untraceable currency. It is open-source and freely available to all. With Monero, you are your own bank. Only 
-                            you control and are responsible for your funds; your accounts 
+                            <p id="main-text">Monero is a secure, private and untraceable cryptocurrency. It is open-source and accessible to all. With Monero, you are your own bank. Only
```
rehrar: Actually this needs an Oxford comma. "Monero is a secure, private, and untraceable cryptocurrency."

index.html
```diff
@@ -40,7 +40,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                     <div class="row middle-xs info-block-row private">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12  why-text">
                             <h3>Monero is secure.</h3>
-                            <p>Monero can't be hacked to steal your funds, due to the power of distributed consensus, which you can read about here. This means that you are responsible for your money, and don't have to trust any entity to keep it safe for you.</p>
+                            <p>Monero cannot be hacked to steal your funds, due to its decentralized nature. Only you are responsible for securing your Monero.</p>
```
rehrar: I agree with the first change, but the second sentence doesn't need to change. That's just my opinion.

index.html
```diff
@@ -52,14 +52,14 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is private.</h3>
-                            <p>The power of the blockchain usually increases security at the cost of privacy, but with Monero's sophisticated privacy-centric technology, you get all of the security benefits of the blockchain without any of the privacy trade-offs.</p>
+                            <p>Security of the blockchain normally increases at the cost of privacy. Monero is based on the Cryptonote application layer protocol, which allows for the security benefits of blockchain technology without any of the privacy trade-offs</p>
```
rehrar: I think it would be better if read: "Blockchain technology usually increases security at the cost of privacy." The rest is fine. You're missing a period at the end here.

index.html
```diff
@@ -52,14 +52,14 @@ <h2 id="main-h2">Private Digital Currency</h2>
...
                         </div>
                         
                     </div>
                     <div class="row middle-xs info-block-row">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is untraceable.</h3>
-                            <p>By taking advantage of ring signatures, Monero makes it ambiguous which funds have been spent, and thus extremely unlikely that a transaction could be linked to particular user.</p>
+                            <p>Sending and receiving addresses as well as transacted amounts are obfuscated. Transactions cannot be linked to a particular user or real-world identity.</p>
```
rehrar: Maybe 'obfuscated by default'?

index.html
```diff
@@ -71,7 +71,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is fungible.</h3>
-                            <p>Because of the on-by-default nature of the privacy technologies, Monero is fungible, meaning one Monero will always be equal to another. This means you won't have to worry about Monero blacklisted by exchanges or vendors.</p>
+                            <p>Privacy is always-on. Therefore Monero is fungible, meaning  Monero will never be blacklisted by exchanges or vendors due to its association in previous transactions.</p>
```
rehrar: We can't use the word 'never'. It was in the original copy, but was dropped because it's an absolute statement that puts a lot of trouble to back up. And I think a slightly longer explanation of what 'fungibility' is (like in the original) is warranted. Again, just my opinion.

Great work! Thanks for working on cleaning this up.

index.html
QuickBASIC: I suggest we pull the updated verbiage out of #314 (_strings_en.yml). I don't like the almost accusatory language of "only you are responsible for your funds" and instead prefer instead suggesting that you have complete control. (Using positive language instead of negative.)
> Monero is a secure, private, untraceable currency that is open-source and freely available for anyone to use. With Monero, you are your own bank; You have complete control over your funds.

index.html
QuickBASIC: > the second sentence doesn't need to change

I think the whole line probably should be rewritten. Maybe:
`Due to it's decentralized nature, Monero cannot be hacked or stolen. The power of distributed consensus means that transactions are immutably recorded on the blockchain and you don't have to trust anyone to keep your Monero safe.`

(Distributed consensus link is gone on the live site... probably should just link to Moneropedia entry of the same name if it exists or it can be written.)

index.html
QuickBASIC: > Monero's sophisticated privacy-centric technology
I think you're on the right track that this line needs to be updated. (It's a bit of puffery IMO.)

Maybe: `Blockchain technology usually increases security at the cost of privacy, but because Monero is built with ring-signatures, it has hidden amounts, destinations, and origins of transactions. That means it gains all the benefits of having a public blockchain, without any of the normal privacy concessions.'

index.html
QuickBASIC: > Maybe 'obfuscated by default'?

I think `private by default` has become somewhat of a tagline for Monero. I think we should use it as much as possible if appropriate.

>Sending and receiving addresses as well as transacted amounts are obfuscated.

I think the word `hidden' is better in this context.

For SEO I think it might be good if we keep the word `Monero` in there somewhere. Maybe something like:
`With Monero, sending and recieving addresses are hidden and amounts are obfuscated. With ring-signatures, Monero makes it's ambiguous who is sending funds, and thus it's extremely unlikely that a transaction could be linked to a particular user or real-world identity.`

index.html
QuickBASIC: > Privacy is always-on.
I think `private by default` has become somewhat of a tagline for Monero. I think we should use it as much as possible if appropriate.

>Therefore Monero is fungible
A leap. Maybe: 
`Because Monero is private by default, every transaction is private. That means that no Monero can be traced back to anything "nefarious". This means that Monero is truly fungible, every single Monero will always be equal to another. This makes it very unlikely that Monero could have be blacklisted.`

tyrionmcmaster:
Updated. Thank you everyone for your input. Please review and comment.

QuickBASIC:
We follow the [Collective Code Construction Contract (C4)](https://rfc.zeromq.org/spec:22/C4/) with the Addendum:
>Development Process
>    Maintainers MUST NOT merge pull requests in less than 168 hours (1 week) unless deemed urgent by the Monero Site Maintainer Team

See the [contribution guidelines for this project](https://github.com/monero-project/monero-site/blob/master/CONTRIBUTING.md).

Fluffy will come along eventually and merge a million things in one sitting. Don't worry.

PR #377, opened 2017-08-29
Added missing wallet-rpc calls

cryptobadger:
I added the missing methods. Also did some minor changes/fixes on the existing doc. Probably better if someone with more experience reviews the changes. Especially for the "sweep_all" method descriptions of the parameters would be nice.

Is there a simple way to get better looking json output on the monero site (for the example results)?

### DIFF
--- a/resources/developer-guides/wallet-rpc.md
+++ b/resources/developer-guides/wallet-rpc.md
@@ -94,15 +94,15 @@ Outputs:
 Example:
 
 {:.cli-code}
-<span style="color: cyan;">[ monero->~ ]$</span> curl -X POST http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"getbalance"}' -H 'Content-Type: application/json'
-{
-  "id": "0",
-  "jsonrpc": "2.0",
-  "result": {
-    "balance": 140000000000,
-    "unlocked_balance": 50000000000
-  }
-}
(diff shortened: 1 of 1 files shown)


QuickBASIC:
Put \``` above and below a code block. 

> \```
>10 PRINT "Fluffyponyza is a real cool guy"
>20 GOTO 10
> \```


I'm not sure the that part of the markdown gets rendered to HTML by the site though. Look in for the kramdown documentation or maybe ask @rehrar.

QuickBASIC:
@cryptobadger Had some time this morning to look it up. It looks like the markdown in use by [Jekyll uses the Reddit style codeblock](https://daringfireball.net/projects/markdown/syntax#precode) paradigm, so you should put 4 spaces before each line of the codeblock. You'd have to remove the \<span\> though because if you put it in a code block I believe it will be ignored.

Like this:

```
    [ monero->~ ]$ curl -X POST http://localhost:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"sweep_all","params":{"address":"9sS8eRU2b5ZbN2FPSrpkiab1bjbHE5XtL6Ti6We3Fhw5aQPudRfVVypjgzKDNkxtvTQSPs122NKggb2mqcqkKSeMNVu59S","mixin":2,"unlock_time":0,"get_tx_keys":true}}' -H 'Content-Type: application/json'
```

cryptobadger:
Thanks for your response. I think i'll leave the curl statement as it is now, because it looks ok and i'll indent the resulting json output. 
Your thoughts about this?

QuickBASIC:
I think indenting it may be better. I was just trying to answer your question.

cryptobadger:
i reformatted the code examples with 4 space indentation. also added a few missing parameters. please review :)

PR #381, opened 2017-09-01
Update "Using Monero" page

tyrionmcmaster:
Please review and comment.

### DIFF
--- a/get-started/using/index.md
+++ b/get-started/using/index.md
@@ -17,7 +17,7 @@ title: Using Monero
                         </div>
                     </div>
                     <div class="row start-xs">
-                        <p>Monero is a secure, private, and untraceable cryptocurrency. The developers and community are committed to protecting these values. Learn more by reading the <a href="/get-started/what-is-monero/">What is Monero page</a>. The <a href="https://github.com/monero-project">source code</a> is also available for review and discussion.</p>
+                        <p>Monero is a secure, private, and untraceable cryptocurrency. The developers and community are committed to protecting these values. Learn more by reading the <a href="/get-started/what-is-monero/">What is Monero</a> page. The <a href="https://github.com/monero-project">source code</a> is also available for review and discussion.</p>
                     </div>
                 </div>
             </div>
--- a/get-started/using/index.md
+++ b/get-started/using/index.md
@@ -92,7 +92,7 @@ title: Using Monero
                         </div>
                     </div>
                     <div class="row start-xs">
-                        <p>Monero may be used to purchase many goods and services. For a listing, see our <a href="/community/merchants/">Merchants</a> page. </p>
+                        <p>Monero may be used to purchase many goods and services. For a listing, see the <a href="/community/merchants/">Merchants</a> page. </p>
                     </div>
                 </div>
             </div>
--- a/get-started/using/index.md
+++ b/get-started/using/index.md
@@ -3,7 +3,7 @@ layout: custom
 title: Using Monero
 ---
 <div class="text-center container description">
-    <p></p>
+    <p>Transacting with Monero can be made easy. This page is designed to guide users in that process.</p>
 </div>
 <div class="using">
     <section class="container">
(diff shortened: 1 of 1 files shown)


get-started/using/index.md
```diff
@@ -3,7 +3,7 @@ layout: custom
 title: Using Monero
 ---
 <div class="text-center container description">
-    <p>Using Monero is easy, and there are a variety of tools and guides available to help newcomers. This page will guide you through the basics.</p>
+    <p></p>
```
erciccione: Why this is left empty? I think a short introduction would be nice

get-started/using/index.md
```diff
@@ -30,23 +30,23 @@ title: Using Monero
...
                 <div class="info-block">
                     <div class="row center-xs">
                         <div class="col">
-                            <h2>3. Getting a Wallet</h2>
+                            <h2>Obtain a Wallet</h2>
                         </div>
                     </div>
                     <div class="row start-xs">
-                        <p>You need somewhere to store physical cash such as your bank, your wallet, or under your mattress. Although Monero is an entirely digital asset (you cannot touch them!), you still need a place to store them. This is done in a digital wallet. You oversee securing your own funds, and there is no way to restore them if you make a mistake. Thus, take great care in selecting the right wallet for you. Some are safer and others are more convenient than others. See our <a href="/downloads/">downloads page</a> for the wallets that The Monero Project has put out.</p>
+                        <p>A Monero wallet is required to secure your own funds. See the <a href="/downloads/">downloads page</a> for a listing of available wallets.</p>
```
QuickBASIC: If we're going to use `Obtain Monero` for the next section I think we should probably pick another verb here.  (`Getting` is fine IMO and `Obtain` sounds a bit stodgy in this context.)

`your own` is redundant. If they're mine (your), then who else would they belong to.

Honestly, I think also think we need to disambiguate from the beginning the difference between a wallet (the software) and a wallet (an account). I think `account` is the preferred term with the developers for this very reason. Maybe:
 
> <p>A Monero wallet is the software that's used secure your funds. See the <a href="/downloads/">downloads page</a> for a listing of available wallets. This will allow you to establish an account that only you control.</p>

get-started/using/index.md
```diff
@@ -59,11 +59,11 @@ title: Using Monero
                 <div class="info-block text-adapt">
                     <div class="row center-xs">
                         <div class="col">
-                            <h2>4. Getting Monero</h2>
+                            <h2>Obtain Monero</h2>
                         </div>
                     </div>
                     <div class="row start-xs">
-                        <p>Empty wallets are no fun! Let’s get some Monero in there! You can receive Monero from someone else, or purchase some on <a href="/community/merchants/">an exchange</a>. We do not endorse specific services, but you can view the exchanges that support Monero on the Merchants page. Depending on the exchange, you can buy Monero from several currencies including USD, EUR, and Bitcoin. For price speculation and talk, join the active trading discussion at <a href="irc://chat.freenode.net/#monero-markets">#monero-markets.</a></p>
+                        <p>One may purchase Monero on an <a href="/community/merchants/">exchange</a> with fiat or other cryptocurrencies.
```
QuickBASIC: > One may

Although `you` and `one` are both considered impersonal/indefinite pronouns we should probably avoid mixing them. Generally `you` is considered more informal/friendly in American English. Also,  `may` connotes permission where `can` permits capability to do so. Perhaps: `You can purchase`.

Also, this might be a great spot to insert something about mining.

get-started/using/index.md
```diff
@@ -76,26 +76,26 @@ title: Using Monero
...
+                        <p>Learn how to send and receive Monero by <a href="/get-started/accepting/">viewing our guide</a>.</p>
                     </div>
                 </div>
             </div>
             <div class="right half col-lg-6 col-md-6 col-sm-12 col-xs-12">
                 <div class="info-block">
                     <div class="row center-xs">
                         <div class="col">
-                            <h2>6. Now What?</h2>
+                            <h2>Transact with Monero</h2>
```
QuickBASIC: `Transact` is a bit impersonal. Perhaps change to `Using`.

PR #385, opened 2017-09-02
Cleanup IRC channels

mattcode55:
* Move IRC channels into their own yml file
* Add a couple more Monero related IRC channels
* Fix the style of the hangouts page to accommodate for the new channels


erciccione:
Can you add the newborn `#monero-translations` ? 
The channel is about localizing Monero in other languages
thanks

PR #396, opened 2017-09-10
touching up the wording on the main page

apezio:
Monero is secure
Do we really want the first sentence about the security of Monero to include "extremely difficult to hack and steal your funds. "  To me that doesn't sound safe for large amounts if it's only "extremely difficult"... and why mention "hack" and "steal" at all?  Further, if you are trying to define "decentralized cryptocurrency" the definition is absolutely NOT "extremely difficult to hack and steal your funds.".  You could say the same of cash in a bank vault, an ATM machine, a centralized currency, or the safe in my basement...  Those are all "extremely difficult to hack and steal" from.  It seems to me we should be explaining what decentralized cryptocurrency is and perhaps state it's the safest method known to mankind to store value, outside of Fort Knox.  Not that its just hard to hack and steal.

Monero is fungible - if we're going to say it's untraceable above, we may as well remove the whole "In its current state, it is extremely unlikely that Monero will ever be..." and just bluntly state that it will never be blacklisted due to previous association.  Also, very few people know what fungible even means, so I have added the definition so people know what they're reading.

### DIFF
--- a/index.html
+++ b/index.html
@@ -153,7 +153,7 @@ Title: Home
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>Moneropedia</h2>
-                                <p class="main-resources">An open resource which demystifies Monero and cryptocurrency.  Here you will find an alphabetical guide to terms and their meanings from both the Monero and Kovri projects.</p>
+                                <p class="main-resources">Would you like to look up the meanings of the terms and concepts used in Monero? Here you will find an alphabetical guide to terms and their meanings from both the Monero and Kovri projects.</p>
                                 <p><a href="/resources/moneropedia/" class="btn-link btn-auto">Read Moneropedia</a></p>
                             </div>
                         </div>
--- a/index.html
+++ b/index.html
@@ -40,7 +40,7 @@ Title: Home
                     <div class="row middle-xs info-block-row private">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12  why-text">
                             <h3>Monero is secure.</h3>
-                                <p>Monero is a decentralized cryptocurrency, meaning secure digital cash powered by an open network of users. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. You are the bank - no one knows how much you have or how you spend it. </p>
+                                <p>Monero is a decentralized cryptocurrency, meaning secure digital cash powered by a network operated by its users. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. You are the bank - no one knows how much you have or how you spend it. </p>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-4 col-xs-12 last-sm first-xs center-xs">
                             <img src="/img/secure-monero.png" alt="Secure safe" class="main-icon">
@@ -71,7 +71,7 @@ Title: Home
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is fungible.</h3>
-                            <p> Monero is fungible because it is private by default. This means no one can tell one Monero from another Monero. Monero will never be blacklisted by exchanges or vendors because there is no association to previous transactions.</p>
+                            <p> Monero is fungible because it is private by default. This means no one can tell one Monero from another Monero and therefore it can't be blacklisted by exchanges or vendors.</p>
                         </div>
                     </div>
                 </div>
@@ -155,7 +155,7 @@ Title: Home
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>Moneropedia</h2>
-                                <p class="main-resources">An open encyclopedia which demystifies Monero and cryptocurrency.  Here you will find an alphabetical guide to terms and their meanings from both the Monero and Kovri projects.</p>
+                                <p class="main-resources">An open resource which demystifies Monero and cryptocurrency.  Here you will find an alphabetical guide to terms and their meanings from both the Monero and Kovri projects.</p>
                                 <p><a href="/resources/moneropedia/" class="btn-link btn-auto">Read Moneropedia</a></p>
                             </div>
                         </div>
@@ -169,7 +169,7 @@ Title: Home
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>User Guides</h2>
-                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet to mnemonic seeds, and even how to edit this website.</p>
+                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet to recovering a wallet, and even how to edit this website.</p>
                                 <p><a href="/resources/user-guides/" class="btn-link btn-auto">Read user guides</a></p>
                             </div>
                         </div>
--- a/index.html
+++ b/index.html
@@ -20,9 +20,7 @@ Title: Home
                         <div class="main-info">
                             <h1 id="main-h1">Monero</h1>
                             <h2 id="main-h2">Private Digital Currency</h2>
-                            <p id="main-text">Monero is a secure, private, and untraceable cryptocurrency. It is open-source and accessible to all. With Monero, you are your own bank. Only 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -169,7 +169,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>User Guides</h2>
-                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet, recovering a wallet, and even how to edit this website.</p>
+                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet to mnemonic seeds, and even how to edit this website.</p>
```
QuickBASIC: This is the landing page for new users to Monero, who may also be newer to crypto, I think we should keep language simple. The moneropedia can use more direct terms.

apezio: Good point, I agree.

index.html
```diff
@@ -40,7 +40,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                     <div class="row middle-xs info-block-row private">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12  why-text">
                             <h3>Monero is secure.</h3>
-                            <p>Monero is a decentralized cryptocurrency, meaning it is extremely difficult to hack and steal your funds. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. Third-parties do not need to be trusted to keep your Monero safe. </p>
+                                <p>Monero is a decentralized cryptocurrency, meaning secure digital cash powered by an open network of users. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. You are the bank - no one knows how much you have or how you spend it. </p>
```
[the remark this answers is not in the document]
apezio: Either way.  "its" not "it's" in this case I think.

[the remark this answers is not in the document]
apezio: The line "Third-parties do not need to be trusted to keep your Monero safe." made me think that have third-parties can hold your Monero even if you dont trust them, and your Monero would be safe anyhow.  Obviously it's the opposite...

index.html
```diff
@@ -155,7 +155,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>Moneropedia</h2>
-                                <p class="main-resources">Would you like to look up the meanings of the terms and concepts used in Monero? Here you will find an alphabetical list of terms and their meanings from both the Monero and Kovri projects.</p>
+                                <p class="main-resources">An open encyclopedia which demystifies Monero and cryptocurrency.  Here you will find an alphabetical guide to terms and their meanings from both the Monero and Kovri projects.</p>
```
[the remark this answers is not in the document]
apezio: I like "resource".  Was thinking encylopedia as in "Moneropedia".  I thought Moneropedia was a play on encyclopedia, but "pedia" itself pertains to learning in general.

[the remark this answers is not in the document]
QuickBASIC: Probably a play on Wikipedia.

index.html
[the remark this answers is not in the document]
QuickBASIC: Sorry on mobile, so wrong `its`. You're right. Yeah, I agree that line needs to be rewritten for clarity. I think the intent of that line is intended to say something about how the parties that contribute to consensus are not privy to your info, so you don't have to trust them.

apezio:
I made the suggested changes.

I also took out the part referencing "mnemonic seeds" as I understand now the intention is to keep things as easy to understand as possible on the main page.  So that edit just replaces a comma with the word "to", which is more grammatically correct.

index.html
```diff
@@ -169,7 +169,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         <div class="row center-xs">
                             <div class="col">
                                 <h2>User Guides</h2>
-                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet to mnemonic seeds, and even how to edit this website.</p>
+                                <p class="main-resources">Step-by-step guides to all things Monero are separated by category and cover everything from creating a wallet to recovering a wallet, and even how to edit this website.</p>
```
QuickBASIC: `creating a wallet to recovering a wallet`

Maybe `create a wallet to setting up a node`, so it shows a dichotomy between easy informational stuff and more technical stuff.

index.html
```diff
@@ -71,7 +71,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is fungible.</h3>
-                            <p> Monero is fungible because it is private by default. In its current state, it is extremely unlikely that Monero will ever be blacklisted by exchanges or vendors due to its association in previous transactions.</p>
+                            <p> Monero is fungible because it is private by default. This means no one can tell one Monero from another Monero and therefore it can't be blacklisted by exchanges or vendors.</p>
```
QuickBASIC: Sorry, if I wasn't clear. Maybe if you clarify that a specific coin can't be blacklisted (because it's funglible)... The wording here isn't clear because it reads as of Monero as a whole can't be blacklisted by an exchange.

index.html
apezio: Yes!  Thats what I wanted to do! :D :D  great change.

apezio: I think 'node' may even be too abstract for the main page, so I used `creating a wallet to supporting the network'.  This hits manys goals; it creates the dichotomy, informs the reader they as an individual can help Monero and likewise Monero is supported by users like them.  

Also it ties into the text from the 'Monero is Secure' section: `... cash powered by a network operated by its users`.

tyrionmcmaster:
`Monero is a decentralized cryptocurrency meaning **it is** secure digital cash powered by a network operated by its users.`

`You are the bank - no one knows how much you have or how you spend it.` 
We already have a statement about this in the top paragraph. I don't believe using the same wording is ideal.

`This means no one can tell one Monero from another Monero and therefore your funds can't be blacklisted by exchanges or vendors.`
I would argue that `due to its association in previous transactions` is a key message of the original statement and that removing it would make the intended message less obvious.

index.html
```diff
@@ -71,7 +69,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                         </div>
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12 why-text">
                             <h3>Monero is fungible.</h3>
-                            <p> Monero is fungible because it is private by default. In its current state, it is extremely unlikely that Monero will ever be blacklisted by exchanges or vendors due to its association in previous transactions.</p>
+                            <p> Monero is fungible because it is private by default. This means no one can tell one Monero from another Monero and therefore your funds can't be blacklisted by exchanges or vendors.</p>
```
tyrionmcmaster: Monero is fungible because it is private by default. Monero will not be blacklisted by exchanges or vendors due to its association in previous transactions.

index.html
```diff
@@ -40,7 +38,7 @@ <h2 id="main-h2">Private Digital Currency</h2>
                     <div class="row middle-xs info-block-row private">
                         <div class="col-lg-6 col-md-6 col-sm-8 col-xs-12  why-text">
                             <h3>Monero is secure.</h3>
-                            <p>Monero is a decentralized cryptocurrency, meaning it is extremely difficult to hack and steal your funds. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. Third-parties do not need to be trusted to keep your Monero safe. </p>
+                                <p>Monero is a decentralized cryptocurrency, meaning secure digital cash powered by a network operated by its users. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. You are the bank - no one knows how much you have or how you spend it. </p>
```
tyrionmcmaster: Monero is a decentralized cryptocurrency, meaning, like Bitcoin, it is extremely difficult to hack and steal your funds. Transactions are confirmed by distributed consensus, and then immutably recorded on the blockchain. Your are your own bank and third parties do not need to be trusted to keep your Monero safe.

apezio:
Removed some of my proposed changes.  Now this commit basically just changes the part about hacking and stealing into a useful description of decentralized crypto.

PR #397, opened 2017-09-10
Fix a FAQ tab

cryptochangements34:
The proper CSS class was changed from "accordian" to "accordion". The "If Monero is so private how do we know they're not being created out of thin air?" section got left out when the change was made which results in the ugly default checkbox being used instead of rehrar's custom CSS


get-started/faq/index.md
```diff
@@ -126,8 +126,8 @@ Fungibility is a simple property of money such that there are no differences bet
 </div>
 
 <div class="tab">
-    <input id="tab-eleven" type="checkbox" name="tabs" class="accordian">
-    <label for="tab-eleven" class="accordian">If Monero is so private how do we know they're not being created out of thin air?</label>
+    <input id="tab-eleven" type="checkbox" name="tabs" class="accordion">
```
MaxXor: This needs to be `tab-twelve` otherwise it will collapse/expand the above tab (below too).

PR #400, opened 2017-09-13
Add noreferrer + noopener attributes to blank targets

anonimal:
Increases privacy by requiring supported browsers to *not* send a
referer header. Also prevents tabnabbing.

### DIFF
--- a/community/hangouts/index.md
+++ b/community/hangouts/index.md
@@ -10,16 +10,16 @@ The Monero community is diverse and varied. We come from all over, but we defini
     <section class="container">   
         <ul class="row center-xs">
             <li>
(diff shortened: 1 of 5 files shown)


anonimal:
@MaxXor AFAICT, tabnabbing works with `same-origin` and `cross-origin` policies, and we would still need `noopener`. What's your solution exactly?

MaxXor:
You are right, I forgot tabnabbing. Not sure if a simple http header can fix it for external URLs.

fluffypony:
Please rebase

anonimal:
@fluffypony rebased.

PR #402, opened 2017-09-14
Fix links to RPC methods in RPC guides

mattcode55:
Most of the links to the methods in the wallet RPC guide are broken.


mattcode55:
I think it's to do with how the markdown compiler generates the ids for the header tags. The setting for kramdown is `auto_ids` which only has a boolean to turn on and off the setting, so I'm not sure how the behavior changed.

mattcode55:
Some of the links are broken in the daemon RPC docs too, I'll amend them to this PR.

PR #404, opened 2017-09-15
Add allforcrypto.com to merchants.yml and alphabetically sort merchants

### LINKED ISSUE #403: add my site to merchant list?
Just wondering if you can add my site https://allforcrypto.com/  to merchants who except monero as payment thanks

mattcode55:
* Fix #403
* Alphabetically sort merchants

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -170,6 +170,8 @@
       url: https://silverround.com/
     - name: Synntech Gaming
       url: https://synntechgaming.com
+    - name: All For Crypto
+      url: https://allforcrypto.com
 - category: Entertainment
   merchants:
     - name: Crypto Games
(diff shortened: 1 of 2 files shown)


_data/merchants.yml
```diff
@@ -170,6 +170,8 @@
       url: https://silverround.com/
     - name: Synntech Gaming
       url: https://synntechgaming.com
+    - name: All For Crypto
```
QuickBASIC: Could we move this up to the top of the section to retain the alphabetical listing?

mattcode55: Sorry, I forgot about that. I think it would be better to sort alphabetically in the liquid template so that the order in the `merchants.yml` has no effect on the ordering on the page. I'll create a PR for that in a few minutes.

QuickBASIC: Awesome. I still think it may be beneficial to keep it ordered as the list grows as it makes it easier to maintain. (Making sure not to duplicate items is easier.)

PR #412, opened 2017-09-22
Add Localmonero to Merchant list

### LINKED ISSUE #410: Addition for Merchants List: LocalMonero.co
Hello, Monero team!

Would be nice to see [LocalMonero.co](https://localmonero.co/) on the exchanges list. 

LocalMonero is a Monero P2P exchange, a Monero's equivalent of LocalBitcoins.com. Unlike centralized cryptocurrency exchanges, you make a trade directly with another person. This makes the process lean and fast, as there is no corporate overhead. You get your Monero instantly. Also, LocalMonero.co can support every payment method its user community supports, making it possible for users without access to traditional banking to also be able to buy or sell Monero. For every transaction, LocalMonero.co offers escrow or transaction service to protect the buyer of Monero.

erciccione:
fix #410

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -28,6 +28,8 @@
       url: https://www.livecoin.net
     - name: LiteBit (Bankwire/GiroPay/iDeal/SOFORT)
       url: https://www.litebit.eu/en/buy/monero
+    - name: LocalMonero (P2P exchange)
+      url: https://localmonero.co/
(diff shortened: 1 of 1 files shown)


localmonero:
@erciccione It would be incorrect to say that we are a decentralized exchanged. We are a P2P exchange.

erciccione:
Ok @localmonero , fixed

PR #417, opened 2017-09-23
Fix merchants URL color

MaxXor:
Currently the merchants URLs are black which is different from all other URLs. This PR changes them to orange as well.

edit: I'm not sure if it's intended to be black.


rehrar:
Shoot. Yes they are intended to be black. If you look at the page when they're all orange, it's almost blinding.

PR #423, opened 2017-09-26
Add disclaimer to merchant list.

QuickBASIC:
* add disclaimer

Over the last month, I've seen several instances (Reddit, Github, IRC, etc) of people who said something along the lines of "it's listed on the Monero site, so it should be okay." I think it's important to add a disclosure to remind users of their responsibility to protect their funds and to think critically about who they do business with.

I'm not certain if this block of text should be at the bottom or the top, suggestions would be helpful. I don't have a jekyll instance set up at the moment, so someone please tell me if the `div` is right and will render properly at the bottom of the page.

### DIFF
--- a/community/merchants/index.md
+++ b/community/merchants/index.md
@@ -26,5 +26,5 @@ Merchants of all kinds have come to value the financial privacy that Monero brin
 </div>
 
 <div markdown="1" class="text-center container description">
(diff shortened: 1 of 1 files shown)


SamsungGalaxyPlayer:
I think this is a good idea.

PR #424, opened 2017-09-27
added link to go (golang) libraries and utilities

parasew:


mattcode55:
Maybe it would be better to link to specific libraries instead of this index?

parasew:
@mattcode55 yes, you are right. i originally wanted to link to the index as there are a few implementations out there. i changed it now to the most recent lib.

parasew:
can we merge this?

mattcode55:
Everything gets merged in a big batch whenever fp has time. Just wait :)

PR #455, opened 2017-10-25
Fix formatting on copyright page

mattcode55:
The content on the copyright page isn't inside a container and looks a bit awkward.

###### Before
Before

###### After
After

### DIFF
--- a/legal/copyright.md
+++ b/legal/copyright.md
@@ -1,6 +1,12 @@
 ---
-layout: custom
-title: Legal
+layout: static_page
+title: "Copyright"
+title-pre-kick: "Copyright "
+title-kick: "Copyright"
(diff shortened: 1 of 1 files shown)


rehrar:
The copyright page has been replaced with the full [Legal page](https://getmonero.org/legal/) which has all Copyright, Privacy, and Terms of Use on it. It would be better for the Copyright page to be removed entirely. I usually try to leave 'deleted' pages on the site for a couple months just in case there is great outcry and the changes need to be reverted.

mattcode55:
Sorry, I didn't realise it was meant to be deleted. I followed a link to it from the forums.

It's safe to delete pages because they'll all be in the Git history if you ever want them back.

PR #456, opened 2017-10-25
Fix image width

MaxXor:
Fix the large MyMonero logo on downloads page. The default width shouldn't be set to 100% of the page width.

mymonero large

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -989,8 +989,6 @@ h1, h2, h3, h4, h5, h6, p {margin: 0; padding: 0;}
 
 img, video {max-width: 100%;}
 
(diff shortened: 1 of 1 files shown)


rehrar:
Thanks for this. We added temporarily for a fix on the new Press Release page and forgot to take it out and replace it with the class in the correct place. Working on several computers has caused some of these errors in the past, and I'm working on streamlining the process. 

Making this fix will break the 'Press Release' page, but I'll add the class.

@luigi1111 ping to merge these fixes ASAP.

luigi1111:
@rehrar which other fix?

PR #457, opened 2017-10-25
Fix link to source code on downloads page

MaxXor:


rehrar:
Yeah, I think this was copied directly from the old site. No issues with it though, but better to have it say 'Monero', I agree.

PR #465, opened 2017-11-02
Added in Tuxcon web + mail hosting

stevesbrain:
Tuxcon accepts Monero as payment for web and mail hosting services


QuickBASIC:
+merchant

PR #454, opened 2017-10-24
Crypto_profi patch - add walletmonero

crypto-profi:
Add walletmonero.com to the list of light wallets.

PR #469, opened 2017-11-06
wallet monero light wallet

walletmonero:
Add walletmonero.com to the list of light wallets.


mattcode55:
Looks like a duplicate of #454 

I'm not sure but the consensus seemed to be that you should add WalletMonero to the merchants page, not to the downloads page.

+duplicate

PR #502, opened 2017-12-08
Update list of miners

mattcode55:
* Remove old miners
* Replace xmr-stak-* with new, unified xmr-stak
* Add XMRig, XMRig AMD and XMRig NVIDIA


mattcode55:
@MaxXor I tried that at first. There are fewer miners now, so it looked 
a bit bare with only a couple of items in each column. :(

PR #503, opened 2017-12-08
Fix missing protocol upgrade tag

MaxXor:
Clicking on the protocol upgrade tag link on the bottom of [this blog post](https://getmonero.org/2017/09/13/september-15-2017-protocol-upgrade-hard-fork.html) results in 404.

edit: Somehow I forgot it in my last PR https://github.com/monero-project/monero-site/pull/500 as this directory was ignored by the `.gitignore` file.

PR #505, opened 2017-12-09
Remove unnecessary entries from .gitignore

MaxXor:
See https://github.com/monero-project/monero-site/pull/503

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ ietemplates/
 _site/*
 .idea/
 *.swp
-tags


.gitignore
```diff
@@ -1,5 +1,3 @@
 ietemplates/
 _site/*
 .idea/
-*.swp
```
mattcode55: Probably best leaving that in, it's vim's swap file. People should put it in `.git/info/exclude` if they use vim, but they never do and end up committing tonnes of .swp files.

MaxXor:
Done!

PR #437, opened 2017-10-10
Add compressed RingCT video

SamsungGalaxyPlayer:
Adds a compressed RingCT video.

Resolves #435.

PR #521, opened 2017-12-18
Add RingCT video to carousel

mattcode55:
Depends on #437


SamsungGalaxyPlayer:
The thumbnail looks good!

PR #524, opened 2017-12-20
Add merchant - monerujo Monero Wallet for Android

hampusohlsson:


MaxXor:
What about linking to https://monerujo.io/ instead of the playstore url?

hampusohlsson:
Good point, I'll update

MaxXor:
Nice, could you squash the two commits into one? Then it's ready to get merged I think.

hampusohlsson:
Fixed

SamsungGalaxyPlayer:
Can't it be added here too? https://getmonero.org/downloads/#mobilelight

hampusohlsson:
Added it to download section too

rehrar:
Thank you for your work, but that reddit thread and this PR could not have come at a worse time. Take the $40 of XMR, I don't want it, but this PR should not be merged due to #527 which adds Monerujo, but in the new format. :)

hampusohlsson:
Haha no worries! It's just awesome that the community is moving forward

luigi1111:
Added in #527

PR #525, opened 2017-12-20
added monerujo to tools

dasmoover:


MaxXor:
#524 was faster, sorry. Also this is in the wrong section and not alphabetically sorted.

PR #526, opened 2017-12-20
Update JetBrains logo

mattcode55:
The JetBrains logo has changed.

Ref: monero-project/monero#2505


MaxXor:
Will the new logo mess up the [site](https://getmonero.org/community/sponsorships/)?

mattcode55:
##### Before
Before

##### After
After

Maybe Araxis and JetBrains could swap positions so that the square logo looks less out of place?

MaxXor:
@mattcode55 That would be nice if you could swap the two logos. :)

PR #537, opened 2018-01-06
Add Chicago Meetup to events

SamsungGalaxyPlayer:


SamsungGalaxyPlayer:
@MaxXor the current website guide says the following:

> and paste it IN THE CHRONOLOGICAL ORDER that it will be in. So if there is an event happening before yours, make sure it is higher up than yours. If there is an event happening AFTER yours, make sure it is after your event.

If @rehrar thinks this is the way to go, I can remove the past event entirely.

rehrar:
I think we should remove the Events page for now. It's too much upkeep for not enough benefit.

rehrar:
@SamsungGalaxyPlayer can we close this as the Events page has been obliterated with extreme prejudice?

PR #546, opened 2018-01-10
Remove tracking parameters from WalletMonero link

mattcode55:
Given the changes people have made to improve the privacy of users on the site (#400, #442), it seems a bit silly to leave tracking codes in links.


mattcode55:
@MaxXor Whoops, didn't notice that one. Looks like there's two sites with the Google Analytics tracking parameters.

PR #556, opened 2018-01-16
Describe "do_not_relay" flag and the strange default

emesik:
This important flag was not documented. What's worse, the default value stops tx propagation, which definitely need mentioning.


rehrar:
@emesik can you rebase this? Then we let's get it merged.

Issue #570, opened 2018-01-23
Tweak messaging in headline paragraph?


robby-dermody:
I generally lurk on the reddits and slack, and I've noticed a bit of a shift with how many are thinking about Monero. Whereas before, its privacy and fungibility aspects were highly emphasized as standalone -- almost abstract -- properties, many seem to be coming around to the more holistic realization that these are required attributes of good digital money first and foremost. I think this is a great development, as an analogous relationship to the _better_ properties of paper cash (specifically, it's accessibility, privacy, bearer instrument nature and fungibility) massively helps people initially grok the value of Monero. In doing so, it may also help deflect the "bad boy" narrative that the press loves to drape all over Monero: that it's a coin used by hackers, drug dealers, and people who are generally up to no good, and that by extension, _you_, not being one of these kinds of people, _don't need it_.

So with all that said, I wanted to get peoples' opinions towards tweaking the first paragraph on the website. You can see the current content by going to [getmonero.org](https://getmonero.org) (it's the next to the right of the embedded video, at the top). I'd like to change it to something like this:

**Monero**
**(tagline:) True Money** _(from the comment [here](https://github.com/monero-project/meta/issues/146) with the highest likes...or, have no tagline)_

**Monero is cash for a connected world. It’s fast, private and secure. With Monero, you are your own bank. You can spend safely, knowing that others cannot see your balances or track your activity.**

Justifications:

* The updated text is shorter and more to the point. It immediately likens Monero to something that virtually everyone intrinsically understands and values: cash.
* With the advent of 0-conf spending with the Kasisto POS, mobile wallets, and monero's shorter blocktimes, I think we have enough justification to say it is "fast". People understand and -- as we've seen with all the interest for things like IOTA, XRB, etc -- _value_ fast.
* I carried over the "private" and "secure" for obvious reasons.
* I kept the bit on "you are your own bank" because it's important, is understandable and is what cash is all about (being a bearer instrument).
[...]

PR #601, opened 2018-02-13
Tweak headline paragraph as per discussion at #570 (take 2)

robby-dermody:
Reopened with the updated wording applied to the new i18n site as I messed up the rebase on the first one. I kept the tagline as-is due to lack of consensus with that. 

For more info, please see #570 and [relevant reddit thread](https://www.reddit.com/r/Monero/comments/7tftjy/id_like_the_communitys_feedback_on_my_proposed/).

I also tried to avoid the diff to the closing </div> line, but the filetypes are showing up as DOS to me in vim (on linux), and converting to UNIX will modify all line endings in the file and produce a mess of a diff.

### DIFF
--- a/_i18n/en/index.md
+++ b/_i18n/en/index.md
@@ -15,7 +15,7 @@
                         <div class="main-info">
                             <h1 id="main-h1">Monero</h1>
                             <h2 id="main-h2">Private Digital Currency</h2>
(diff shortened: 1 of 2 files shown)


robby-dermody:
@luigi1111 added the oxford comma. let me know if you have any other comments or if this looks good. :)

robby-dermody:
@luigi1111: thank you! Looking forward to it live on the site

PR #628, opened 2018-03-01
Added link to original btctalk thread in all languages

goktugyil:


goktugyil:
Technically yes. Practically why does that matter? They can all be reviewed/merged with a single click.

luigi1111:
@goktugyil It keeps the history clean and somewhat logical (you have gratuitous merges in particular that muddy things up).

PR #634, opened 2018-03-03
Update Core team blog post

anonimal:
Referencing #630.

Now begins a true test of the C4 :smile:


_posts/2018-03-01-core-team-announcement.md
```diff
@@ -8,21 +8,20 @@ author: The Core Team
...
+The role of the Core Team can often be a confusing topic, especially with the various leadership structures employed by other cryptocurrencies. As the technology for a completely trustless society is not yet a reality, all projects, even decentralized ones, benefit from having individuals that can carry out tasks that require trust. A quick example of this is owning a domain name (like getmonero.org) or maintaining servers (on which getmonero.org is hosted). Ultimately, there is not yet a way for a domain name to be owned by a decentralized community without using a mixnet, so someone has to be the name holding this domain. It's in scenarios like this where the Core Team are able to work on bridging the decentralized world we are trying to bring about with the centralized reality in which we are currently living.
```
rehrar: I think this an excellent clarification!

_posts/2018-03-01-core-team-announcement.md
```diff
@@ -8,21 +8,20 @@ author: The Core Team
...
 - Act as trusted signers and distributors of reference clients for the Monero coin, and other related technologies.
-- Set a direction and vision for the Monero Project
     
 In the event that the community deems the Core Team unworthy or lax in their duties, they are free to fork the project away from the current Core Team, as happened with thankful_for_today.
 
 There are some things that the Core Team are NOT responsible for, despite misconceptions from the community. Some of the more common misconceptions are listed below:
 - Members of the Core Team are NOT anyone's boss, and no permission is needed from them to do anything.
 - The Core Team does NOT act as a centralized point of failure, but encourages organic, self-started initiatives that further the ecosystem of Monero.
-- The Core Team does NOT equal Monero. In the event that one, or all of the Core Team goes rogue, we are to remember that Monero is a movement. A global initiative to further privacy globally, and provide real, fungible, digital money for everyone. This can happen even without the presence of the Core Team.
```
rehrar: Not sure I agree that a movement is only only defined as by having leaders. I believe a decentralized movement is possible, at least under a definition of a group of people with similar ideals working toward a changed future, even without explicit leadership structures. 

As well, I don't think the wording of 'global' is trying to signify globalism as much as highlight the fact that the work and ideals of Monero are not bound to a particular country or place.

Perhaps a better clarification would be that the "Monero Project" is an initiative. Because it's true that Monero Is a software (a suite of them actually), but the Monero Project has multiple softwares and other initiatives under its umbrella. 

In this sense, it can be said that "one of the goals" is to provide "real, fungible, digital money for everyone", seeing as how that's not the only goal.

anonimal: >a movement is only only defined as by having leaders

I don't think I said this. It's not *only* defined.

Every movement has leader(s) and authorities behind it regardless of whether they are immortalized figureheads or not. In "movements", someone has to push the button, someone has to write the pay checks, someone has to say "let's do this at this time" so the "movement" moves.

As noted in https://github.com/monero-project/monero-site/pull/630#discussion_r171994026:

>at least under a definition of a group of people with similar ideals working toward a changed future

You're describing a cabal. Cabals are centralized bodies. Also see https://github.com/monero-project/monero-site/pull/630#discussion_r171994026.

>explicit leadership structures

That's the point of "movements": to hide those who fund them. Implicitness is a great disguise.

People don't simply act on their own in groups; they are like fish, like sheep, and whether explicitly defined or implicitly defined, there are authorities within groups which dictate the movement of the group. Dictating can be as simple as the power of suggestion.

>the work and ideals of Monero are not bound to a particular country or place.

Nor is the international banking system. And how is that a bragging right?

>Perhaps a better clarification would be that the "Monero Project" is an initiative

An initiator implies authority which in turns implies followers. This creates centralization as noted in https://github.com/monero-project/monero-site/pull/630#discussion_r171994026.

>for everyone

...for everyone who uses a computer. That's not everyone. Monero is not for Amazonian tribesmen unless they want to sacrifice their lifestyle for the sake of the electron.

_posts/2018-03-01-core-team-announcement.md
anonimal: This is making me think about a devising a new multi-homing scheme.

_posts/2018-03-01-core-team-announcement.md
rehrar: > Interestingly, one could even go as far as to say that monero is separate from Monero Project because the software dictates the project more often than the project dictates the software. I'm not opening that can of worms, just an observation

Indeed. I'm a fan of renaming 'The Monero Project' to something else for clarity since it's more than just Monero. :)

_posts/2018-03-01-core-team-announcement.md
```diff
@@ -31,4 +30,4 @@ But every change brings new opportunity, and the Core Team is delighted to welco
 All members of the Core Team, past and present want to thank all volunteers and contributors for their consistent and tireless work. We look forward to what 2018 has in store for us as a project, and as a community.
 
 ![Monero Core Team Cartoon](/img/monero-core.jpg)
-Image credit: [helloluis](https://cryptopop.net/)
\ No newline at end of file
+Image credit: [helloluis](https://cryptopop.net/)
```
luigi1111: Why does this show as changed?

mattcode55: It added a newline to the end of the file.

anonimal: >Why does this show as changed?

I used vim on linux and the file was most likely created in windows. I can try to revert if needed.

_posts/2018-03-01-core-team-announcement.md
anonimal: But it's *not* more than just Monero. The Monero Project's sole purpose is to maintain a certain state of monero software in a way that the project maintainers see fit. How end-users, contributors, developers, and researchers *feel* about the software is *supposed* to be outside the scope of the project - and for the most part is. In fact, the more de-tangled the project is from the software; the better.

Fortunately, the Monero Project is *not* monero software. All monero software and sub-projects can exist without the Monero Project - but the project(s) can't exist without monero. This dichotomy creates a resilient ecosystem by forcing the hand of the maintainers to pursue a *less is more* approach (and also makes Monero Project a fascinating project to contribute to).

If people want to attach a paradigm to the code, so be it. Plenty have done that in the past but ultimately the quality and usefulness of the code outlives anything else.

anonimal:
>Now begins a true test of the C4

Tick tock.

anonimal:
Also see notes regarding signing-off: https://github.com/monero-project/monero-site/pull/630#issuecomment-384101352

el00ruobuob:
@rehrar would you like me to test this one?

rehrar:
This is an update to a blog post, so it shouldn't be necessary.

PR #644, opened 2018-03-08
Alphabetize user guides page.

qubenix:
I noticed the lack of order when I was working on some contributions. Better to do this now before it gets too out of hand.


mattcode55:
You might want to do something like [what the merchants list does](https://github.com/monero-project/monero-site/pull/404/commits/e982768276ec705640b160001da49865cf5a77bf), sorting the guides in the template.

PR #659, opened 2018-03-23
Error in hangout.md files (wrong link to StackExchange)

el00ruobuob:
Fix link to hiding tcp miner trafic incorectly set to the crypto primitives & concepts in the stackexchange questions and answers


el00ruobuob:
I messed up so much time to make a clean pull-request on that... i should have learned more about gitHub first. But now, it's ok.
So i found this error while translating to french, and fixed it for all languages and the template.

PR #661, opened 2018-03-27
Replaced tacotime with binaryFate on the team page

rehrar:


stoffu:
Maybe move tacotime to the Special Thanks section in his honor?

rehrar:
Done @stoffu

PR #676, opened 2018-04-09
Removed Events from the navigation so as not to be an idiot

rehrar:


el00ruobuob:
Question: Should i remove Event page and link from my French PR #673 in a new commit? In a squashed commit? Or later?

PR #691, opened 2018-04-14
Add cryptocurrencytools.org

cryptocurrencytoolsdotorg:
Please consider adding cryptocurrencytools.org as a new offline javascript tool.
It works like moneroaddress.org but tries to make multisig as easy as possible so it gets used more.


jtgrassie:
This should be removed until the source code can be vetted (which currently it is near impossible to).

erciccione:
@jtgrassie none of the services on the "merchants" page have been vetted

jtgrassie:
@erciccione fair enough.

PR #705, opened 2018-04-27
Developer-guides, "Atomic units" replacement with "Piconero(j)"

el00ruobuob:
After translating the deamon developer-guide and reading about the denomination, i figured out we were refering to "atomic units" which are not defined anywhere else.

So i suggest to change this and use the official denomination of piconero(j).

This PR include changes in all files for all languages, and an additional link to the moneropedia denomination page for the "heading note" about it.

I even changed it for Polski, but this will need a quick review from @zofiazinha to make sure something is not inconsistent.


el00ruobuob:
In such a case, i believe i have to update moneropedia articles: create an atomic unit entry and cross-refer denomination's piconero and it as the actual equivalent?

el00ruobuob:
Done. Let me know if we close this one.

el00ruobuob:
Closed as we will implement #706 instead.

PR #707, opened 2018-04-27
Moneropedia transaction mixin cleanup

el00ruobuob:
While reading moneropedia articles, i found that the Transaction article was still refering to a 'mixin_count' parameter.
I suggest to replace it with a link to \@ring-size article, and adding the ~~faxt~~ fact that it is not only the count of foreign transaction, but ~~also yours~~ all outputs.


el00ruobuob:
Wait, some inconsistency here. I'll commit & squash.

el00ruobuob:
Corrected. The Moneropedia article on the site was containing both "mixin_count" and a "see \@ring-size", along with a "double offline" typo.
I presume "offline, offline" is not twice more secure than "offline"

PR #724, opened 2018-05-06
Update importing_blockchain.md

jcklasseter:
With the most recent gui release download for windows 10 from https://getmonero.org/downloads/ (Lithium Luna) the supported relevant command line option is:
```
 --guard-against-pwnage arg (=1)       Verify blocks and transactions during
                                        import (only disable if you exported
                                        the file yourself)
```

Since --verify 1 is not only not supported anymore, but is default behavior it is misleading and should be removed from the guide.

### DIFF
--- a/_i18n/en/resources/user-guides/importing_blockchain.md
+++ b/_i18n/en/resources/user-guides/importing_blockchain.md
@@ -40,11 +40,11 @@ If your Monero wallet is on another drive you can use `DriveLetter:` for example
 
 Now type in your command prompt window:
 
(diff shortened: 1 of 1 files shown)


MaxXor:
I really suggest we should change the way multi-language is implemented. We should check how https://github.com/bitcoinxt/website/pull/51 have implemented it. It's way easier to maintain.

erciccione:
I agree @MaxXor, updating the website is a pain at the moment. A lot of manual work is needed with every update, that's going to be worse with new languages coming. Maybe @rehrar can take a look at that

MaxXor:
I'll open a new issue later to track progress.

Issue #725, opened 2018-05-07
Simplify multi-language support


MaxXor:
The current multi-language support is hard to maintain and gets out of sync very fast. A good example how easy it can be implemented in Jekyll can be seen in https://github.com/bitcoinxt/website/pull/51. I propose that we adapt this to make it less error-prone to update or add pages to the website.

They use `jekyll-multiple-languages-plugin` and have one layout for all languages which contain placeholders like `{% t page.title %}`. Beside that the language files are simple `.yml`-files which contain the translated strings.

This issue can be used to track progress on simplifying multi-language support.

el00ruobuob:
I'm totally in it.
But hasn't @rehrar tried to go this way already? And found a lot of issues / complications?

MaxXor:
@el00ruobuob He mentioned it a while ago, I don't remember the exact details though... Well, I'm sure we can find solutions for it.

MaxXor:
@rehrar Nonetheless I think this is the right decision to do. I think we can fix the languages which read right-to-left later, maybe even with the plugin authors. Thanks for your work, I'll try to help where I can.

rehrar:
After looking at the site, the portions of the site that are not broken up into snippets (such as Guides) should remain as they are, as this will ultimately be the easier translating experience. All the other sectioned portions can be changed over to the new way.

rehrar:
It occurs to me that with this method, it becomes much more difficult to put the untranslated snippet in the pages. The website is actually quite complex with all of the things it has, and I'll need to think about how to do all of the little teeny things.

MaxXor:
Untranslated snippets should have the default value copied from the english translation. We can split up the work and everyone who wants to help does a few pages.

MaxXor:
Maybe just remove it altogether and just display the page in english.

rehrar:
From a UX perspective it's pretty important. It can confuse a user who thinks they may have navigated to the wrong page by accident. I'll see if I can't make something work with the new way of doing things somehow.

el00ruobuob:
@rehrar why not keeping/adding a localized page with only the snippet and an included content then?
```markdown
{% include untranslated.html %}
{% include content.html %}
```
This md file will be just a bridge between html and localization to include the snippet.

el00ruobuob:
Great minds think alike @rehrar.
I wonder if we should add a 
```yml
translated: yes / no / partially
```
[...]

PR #739, opened 2018-05-18
[POC] DO NOT MERGE - proof of concept of translation optimization

el00ruobuob:
This is a POC of translation optilization as discussed in #725
The download page is changed using the usual yml file, a generic html file, and a md file which only includes the HTML along with the snippet for untranslated pages.

@rehrar, please have a look on my publicly available local build on which you know how to connect.


el00ruobuob:
Great! I wonder if we should add a 
```yml
translated: yes / no / partially
```
Statement and use it with a `{% if ... elseif ... %}` to display the message, even a different message for partial translation, and moreover tell the world not to touch those files, and only change the yml.

el00ruobuob:
As @rehrar is already working on the exact same stuff, i'm closing it.
All discussions may be done in #725

PR #754, opened 2018-06-04
README: edit instructions for translators

erciccione:
@rehrar looks like it only needed few tweaks in the README. Did I miss something here? Waiting your review then i will update the guide on Taiga as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -502,8 +502,8 @@ Navigate to the `_data/lang` folder and copy the `template` folder. Paste it int
 Translate the content of the files. Do not touch anything labeled `url`, and in the roadmap.yml ONLY translate the `name:` content.
 
 ### 14.3 _i18n folder
-Navigate to the _i18n folder and duplicate the template.yml file. Rename the duplicate to the two letter language code of your language with a `.yml` at the end. Repeat this process with the `template` folder as well.
-**The template folder and yml file themselves should still be there. They should not be renamed. There should be a new folder and yml file in addition to the ones that were already there.**
+Navigate to the _i18n folder and duplicate the en.yml file. Rename the duplicate to the two letter language code of your language with a `.yml` at the end and change all sections marked 'translated: "yes"' to 'translated: "no"'. Change back to 'yes' once you have translated that section. Now duplicate the `en` folder and rename it with the correct language code.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -502,8 +502,8 @@ Navigate to the `_data/lang` folder and copy the `template` folder. Paste it int
 Translate the content of the files. Do not touch anything labeled `url`, and in the roadmap.yml ONLY translate the `name:` content.
 
 ### 14.3 _i18n folder
-Navigate to the _i18n folder and duplicate the template.yml file. Rename the duplicate to the two letter language code of your language with a `.yml` at the end. Repeat this process with the `template` folder as well.
-**The template folder and yml file themselves should still be there. They should not be renamed. There should be a new folder and yml file in addition to the ones that were already there.**
+Navigate to the _i18n folder and duplicate the en.yml file. Rename the duplicate to the two letter language code of your language with a `.yml` at the end. Repeat this process with the `en` folder as well.
```
el00ruobuob: duplicated en.yml should be updated:
- set all sections to untranslated (translated = "no")
- remove all english sentences so that Jekyll will automatically pick up default (English) one.

Maybe a template.yml make sense here.

README.md
```diff
@@ -502,8 +502,8 @@ Navigate to the `_data/lang` folder and copy the `template` folder. Paste it int
...
+**The original folder and yml file themselves should still be there. They should not be renamed. There should be a new folder and yml file in addition to the ones that were already there.**
 
 Enter the .yml file and translate everything there.
```
el00ruobuob: Set the "translated" variable of each section to "yes" when done

README.md
erciccione: updated with your suggestions. And yes, i agree a template here would be useful, even just to avoid to change all those 'yes' to 'no' every time

el00ruobuob:
Are you adding the template.yml or not after all?

erciccione:
i think last word on this should go to @rehrar

el00ruobuob:
Then I wait for @rehrar to approve!

el00ruobuob:
Is @rehrar lost in translation (reimplementation for kovri)?

rehrar:
This is fine. I think the verbiage of the README should be updated to have the bigger emphasis on the fact that .yml files are a bigger thing, but that Guides still need to be done in the folders. Meh. This is good. Who cares?

Issue #761, opened 2018-06-10
Change version structure of downloads page


dEBRUYNE-1:
In my opinion, the current structure of the downloads page is quite misleading. It may work properly if both versions match, but not in situations where the CLI version is ahead of the GUI version (a situation we're currently experiencing). 

I propose to list the versions for both the CLI and GUI separately to avoid any confusion.

el00ruobuob:
Do you mean we should state of a latest cli version vs. Latest gui version?
So a listing per type, Cli first then Gui, should be more appropriate to a listing per os.

wowee2:
Thank you finally ..so its not just me.

I was under the impression that the GUI was already made since the site says Under Linux64
Current Version: 0.12.2.0 Lithium Luna
which is misleading if CLI is made but GUI is not made.

Please label version on top of each OS version
https://github.com/monero-project/monero/issues/4044

wowee2:
For example see here
![monero gui](https://user-images.githubusercontent.com/40521956/41811079-f0b4f5dc-7708-11e8-8d51-4f1b2ceffb0d.jpg)

el00ruobuob:
Already solved by @rehrar in #766 @wowee2 could you close @dEBRUYNE-1 or @erciccione ?

erciccione:
+resolved

PR #766, opened 2018-06-15
Changed GUI and CLI to have their own versions

rehrar:
- downloads.yml version tag has been separated to two, one for gui and one for cli
- Downloads page has been adjusted to reflect the changes


rehrar:
Solves #761

PR #776, opened 2018-06-23
Logs from 2018-06-23 Community meeting

el00ruobuob:
Tested on the following platforms:

  | Edge | IE 11 | Chrome | Firefox | Opera | Vivaldi
-- | -- | -- | -- | -- | -- | --
Windows 7 | - | OK | OK | OK | OK | OK
Windows 10 | KO | OK | OK | OK | OK | OK
Ubuntu 18.04 | - | - | OK | OK | OK | OK
CentOS 7 | - | - | - | OK | OK | OK

Edge is displaying the content box larger than it should.
It is because of long link `https://taiga.getmonero.org/media/attachments/5/7/4/5/7d6f57b35ee6dc6df9f74534cf9a3693a3d04741d0c97cfca2837692fcda/quickfacts_v2_g.pdf` not been split, so perhaps CSS should be corrected here. Any clue @rehrar?
Should i split the link in two to avoid this mess?

### DIFF
--- a/_posts/2018-06-23-logs-for-the-Community-meeting-held-on-2018-06-23.md
+++ b/_posts/2018-06-23-logs-for-the-Community-meeting-held-on-2018-06-23.md
@@ -1,0 +1,279 @@
+---
+layout: post
+title: Logs for the Community Meeting Held on 2018-06-23
+summary: Community highlights, Community workgroup developments, Forum Funding System updates, Localization workgroup report, Outreach workgroup report, and miscellaneous
+tags: [community, crypto]
+author: el00ruobuob / SamsungGalaxyPlayer
+---
+
+# Logs  
+
+**\<sgp\_>** Meeting time!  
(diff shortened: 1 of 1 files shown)


el00ruobuob:
After sleeping on it i found CSS property `word-break: break-word;` was badly handled on MS Edge.
This is solved by using the `word-break: break-all;` instead.
However, chrome does not handle `word-break: break-all;` so i added both. When both are handled by browser `word-break: break-all;` is prefered.

el00ruobuob:
Damned. I did over squashing here.
Not a problem. I'll correct it. I think about something for the Edge behavior that I'll have to try.

el00ruobuob:
I cannot manage to get it right on Edge without breaking anything.
As some other posts are in the same condition, like [this one](https://getmonero.org/2018/05/29/logs-for-the-kovri-defcon-meeting-held-on-2018-05-29.html), i'm giving up for the moment.
Ready to merge, but a problem on posts with long links is to correct on MS Edge.

PR #788, opened 2018-07-03
Forgot a curly bracket :^)

TheQueenIsDead:


TheQueenIsDead:
@MaxXor unfortunately I'm not too good at French, but I'll give it a go

MaxXor:
@TheQueenIsDead You don't need to understand the language just to add a curly bracket to the example.

Issue #735, opened 2018-05-13
Localized user-guides not using local png folders


el00ruobuob:
I was testing the #734 i made to partially close #729 and figured out the screenshots were not updated.
It was because i only updated the /_i18n/XX/ressources subfolder and not the /ressource subfolders.

So it means, our screenshots, localized or not, are not beeing used in the user guides.

That's too bad, because it make sense to have the screenshot in the right language.
I.e., i re-did most of the screenshots of #649 in #697 as the Windows UIs are displaying differently in french.

Is it a known issue @erciccione & @rehrar?

el00ruobuob:
As a woraround, i believe you are copying the png folders each time you build the site? Couldn't it be automated by Jekyll, or any ruby script?

el00ruobuob:
@rehrar & @luigi1111 The current build of getmonero.org is not using the localized screenshot.
Jekyll is not using the _i18n/***/png folders by himself.
Could you apply the workaround above so that readers will have their localized version of screenshots?

el00ruobuob:
@erciccione its +improvement + in progress through #798

erciccione:
+improvement
+in progress

el00ruobuob:
This could be closed

el00ruobuob:
Fixed

Issue #729, opened 2018-05-08
Update user guides with 0.12.0.0 GUI screenshots


SamsungGalaxyPlayer:
Several guides should be updated for the latest 0.12.0.0 release.

The following need their screenshots updated:

https://getmonero.org/resources/user-guides/solo_mine_GUI.html
https://getmonero.org/resources/user-guides/remote_node_gui.html

The following need to have a section for using the GUI, including screenshots:

https://getmonero.org/resources/user-guides/view_only.html
https://getmonero.org/resources/user-guides/prove-payment.html
https://getmonero.org/resources/user-guides/restore_from_keys.html

erciccione:
Maybe we can just point from those articles to the guide instead of create a dedicated part for the GUI? See https://github.com/monero-ecosystem/monero-GUI-guide/blob/master/monero-GUI-guide.md

+improvement

SamsungGalaxyPlayer:
@erciccione I suppose it depends on what sort of resource Monero should provide.

If we want to have references but use StackExchange for most things, then we can simply reference this guide.

If we want to replicate a lot of StackExchange content including common questions, I recommend we take components of the GUI guide (which is excellent btw) and include the relevant parts in these website guides.

el00ruobuob:
Personnaly i think we should update user-guides (in a best effort way) AND add a "more official" user-guide for the GUI, based on the gui guide which must be updated (and localized) in a short timeframe (a week or two) after a new version is released.
In this way, could we enhance the screenshots from the GUI guides? I mean, we could have "gimp layers" with the numbered bullet, so we could easily localized the screenshots. otherwize have a "screenshot toolkit" with all the things ready to be pasted on new screenshots, so everything is coherent across all languages and versions.

el00ruobuob:
View-only is ready on my side.
I'll go with restore from keys now.

Regarding the prove payments, i'll need to run as testnet or a stagenet node and make transactions between wallets to be able to document it.

el00ruobuob:
There's also https://getmonero.org/resources/user-guides/restore_account.html to improve with the GUI part.
And https://getmonero.org/resources/user-guides/restore_from_keys.html should be moved to the Recovery section of user-guides, along with the "restore account" one.

el00ruobuob:
Done with restore from keyx on my side. I'll continue with restore account, as i only have one screenshot to change.

el00ruobuob:
I'm now private-testnet rich and stagenet rich.
[...]

PR #798, opened 2018-07-10
Localized user-guide screenshots

el00ruobuob:
I did notice in #735 that user-guides did not used their local png folders.
So i made this small ruby script to copy localized png files to the site directory during the build.

Working perfectly here.

Then all the localized screenshot i took to solve #729 could be used, and every user could have a localized experience up to the screenshot itself.

Up to you to test @erciccione

### DIFF
--- a/_plugins/screenshots.rb
+++ b/_plugins/screenshots.rb
@@ -1,0 +1,52 @@
+# Copyright (c) 2014-2018, The Monero Project
+#
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without modification, are
+# permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


el00ruobuob:
example in verification-windows-beginner.html:
image
example in restore_from_keys.html:
image

el00ruobuob:
Just change the code to avoid reading language from config file and instead use the "lang" defines in site.config. This way it is more trivial, and the files are copied for all languages each time a language is build.

el00ruobuob:
No one wants to have a look at it? @erciccione @MaxXor @mattcode55 ? @rehrar ?

Issue #805, opened 2018-07-22
Add Bulletproofs to roadmap


Onefox:
we should add something about the bulletproof status to the roadmap over here:
https://getmonero.org/resources/roadmap/

The reviews are going good and the plan as fare as I know is to enable it in the next protocol upgrade.

el00ruobuob:
In fact it is "More efficient range proofs for RingCT to reduce transaction sizes" but I reckon it's not really obvious.

SamsungGalaxyPlayer:
I believe I made that sentence @el00ruobuob before the bulletproof paper was released. It can be changed to mention bulletproofs.

Onefox:
Also it should be marked as Ongoing task.

SamsungGalaxyPlayer:
@Onefox can you take the initiative on this? It should be trivial to update, and it should be able to be completed entirely from within the browser.

el00ruobuob:
And off topic there is and update to push for kovri alpha version on 8/1 and upcoming beta.

el00ruobuob:
This could be closed since #806 is merged. @erciccione?

erciccione:
+resolved

PR #806, opened 2018-07-25
Roadmap update to match bulletproofs & kovri updates

el00ruobuob:
This solves #805
All languages needs to be reviewed (or retranslated) @erciccione.

This should not be merged before 8/1 as it states kovri alpha release is completed at this date.
Is it still the goal @anonimal?


el00ruobuob:
As Kovri alpha has been released in time, this could safely be merge now. Call for reviewers @mattcode55 @MaxXor @erciccione

PR #802, opened 2018-07-20
Update arabian localization with latest merged

el00ruobuob:
Since #741 #756 and #767 were merged, a few things have to been added.

There are new variables to translate in ar.yml @3b7ameed

PR #772, opened 2018-06-22
Library page & content

el00ruobuob:
This is a new Library page, link from ressources and footer.
Accessed through `https://getmonero.org/library/`
Books and other ressources accessibles from `https://getmonero.org/library/<filename>`

It includes a new section in yaml localized (_i18n) file, with list of publications categories (books, magazines), each categories including list of publications. @erciccione, please give your thoughts if any.

Actual Books & magazines published are:
- Books
  - Preview of Mastering Monero (all credit goes to @serhack)
  - Zero to Monero (all credit goes to @UkoeHB)
- Magazines
  - Revuo Monero Q3 2017 (all credit goes to @rehrar)
  - Revuo Monero Q4 2017 (all credit goes to @rehrar)

This replace & enhance #769, discussed in #771, with agreement of @UkoeHB.

Tested OK on the following platforms:

  | Edge | IE 11 | Chrome | Firefox | Opera | Vivaldi
-- | -- | -- | -- | -- | -- | --
Windows 7 | - | OK | OK | OK | OK | OK
Windows 10 | OK | OK | OK | OK | OK | OK
Ubuntu 18.04 | - | - | OK | OK | OK | OK
CentOS 7 | - | - | - | OK | OK | OK

PR #807, opened 2018-07-25
Arabic yml stuff missing (Library + moneropedia lang_tag)

el00ruobuob:
This complete #802, i did not see it was missing as #772 was merged a long time ago.


el00ruobuob:
@erciccione I just realize i forgot to add the language tag to ar yml in moneropedia localization. So i just did it here.

Issue #803, opened 2018-07-21
Request to add BITEXBOOK crypto exchange to the merchants page.


cvnghv:
Dear Monero Team,

Can you add BITEXBOOK crypto exchange to the merchant list as XMR/BTC pair has been recently added?
Here is the link: https://bitexbook.com/trading/xmrbtc

Please, feel free to ask for any additional information if necessary.

Thank you!

cvnghv:
Dear Monero Team, 

There is an answer from you for a long time. 

Is something wrong?

el00ruobuob:
Sorry to have let this behind.
I'll take care of your request.

el00ruobuob:
Done.

cvnghv:
Thank you!

erciccione:
Totally missed this one, sorry @cvnghv 

+merchant
+in progress

erciccione:
+resolved

PR #809, opened 2018-07-27
New exchange Bitexbook

el00ruobuob:
Add bitexbook to solve #803


cvnghv:
Oh, and one more request: can you add it as BITEXBOOK (in capital letters) exactly?

Thank you!

PR #820, opened 2018-07-31
Wikipedia link corrected in Moneropedia articles

el00ruobuob:
While translating Moneropedia, i found out that the link to wikipedia `/etc/hosts` page was wrong.
The page is not under `https://en.wikipedia.org/wiki/Etc/hosts` anymore but under `https://en.wikipedia.org/wiki/Hosts_(file)`

This correct the link for all occurence of it in all moneropedia article of all languages:

- cannonicaly-unique host
- locally-unique host
- subscription


el00ruobuob:
Tested ok.

PR #825, opened 2018-08-08
Updated user guide 'mine-to-pool' to use a CPU miner compatiable with CN7

Ehhoe:
Changed sections specific to the deprecated xmr-stak-cpu miner to be for XMRig miner. Also added section advising against mining on laptops and unprofitable hardware & minor troubleshooting advise.


el00ruobuob:
I'm not sure about this.
Latest version of xmr-stak is compatible with CryptoNight Variant 1. Why do we need to fully change this guide to another miner?
I do agree the guide needs a refresh, at least to reflect the new settings and the automatic config which is now in place with xmr-stak, but i don't see a real need to move it to xmrig.

I suggest you add all this as a whole new guide, and we try to update the xmr-stak-cpu to latest xmr-stak in the next few month.

What do you think about it @Ehhoe @mattcode55 @MaxXor @erciccione @SamsungGalaxyPlayer ?

PR #826, opened 2018-08-08
Proof-reading user-guide/ledger-wallet-cli.md word "original" fix

lh1008:
Hi @el00ruobuob,

Found another word **(original)** to fix in the `ledger-wallet-cli.md` file. Made the fix in the files that aren't translated yet.

:)


el00ruobuob:
Good catch! Thanks @lh1008

PR #827, opened 2018-08-09
Changed 'flag' to 'option' for consistency with CLI'

ordtrogen:
See how the CLI talks about "options" instead of "flags".

image

### DIFF
--- a/_i18n/en/resources/user-guides/ledger-wallet-cli.md
+++ b/_i18n/en/resources/user-guides/ledger-wallet-cli.md
@@ -145,19 +145,19 @@ Congratulations, you can now use your Ledger Monero wallet in conjunction with t
 
 ### 4. A few final notes
 
-1. We'd strongly advise to test the full process first. That is, send a small amount to the wallet and subsequently restore it (using aforementioned guide) to verify that you can recover the wallet. Note that, upon recreating / restoring the wallet, you ought to append the `--restore-height` flag (with a block height before the height of your first transaction to the wallet) to the command in step 3 (Windows), step 5 (Mac OS X), or step 3 (Linux). More information about the restore height and how to approximate it can be found [here](https://monero.stackexchange.com/questions/7581/what-is-the-relevance-of-the-restore-height).
+1. We'd strongly advise to test the full process first. That is, send a small amount to the wallet and subsequently restore it (using aforementioned guide) to verify that you can recover the wallet. Note that, upon recreating / restoring the wallet, you ought to append the `--restore-height` option (with a block height before the height of your first transaction to the wallet) to the command in step 3 (Windows), step 5 (Mac OS X), or step 3 (Linux). More information about the restore height and how to approximate it can be found [here](https://monero.stackexchange.com/questions/7581/what-is-the-relevance-of-the-restore-height).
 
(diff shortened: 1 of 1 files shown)


ordtrogen:
I hear you, but this PR isn't about the Spanish or any other translation so it's better discussed here.

"--daemon-address" is one of the "command line options" you can pass to monero-wallet-cli, that's why I claim "flag" should be "option" in the source. See:

https://en.wikipedia.org/wiki/Command-line_interface#Command-line_option

"option" in this sense can alse be called "flag" or "switch", like the wikipedia article says, but since the help text of monero-wallet-cli uses the term "option", the documentation should use "option" also.

lh1008:
Wikipedia sometimes is a terrible source of information. Take a look at this [wikipedia flag Spanish definition](https://es.wikipedia.org/wiki/Flag).

BlackLotus64:
I can just tell that flag is often used as _bandera_ in spanish but it's more understandable as _marcador_ (like option by the way). In the end it's just an option, flag is more like a technicality used in programmng and computer-related things.

In short, I think it can be changed, because even monerod uses Option, and like I said, flag in the end is just and option.

lh1008:
Let me give it one more try, because I think you aren't getting the context of what "flag", "marcador" in Spanish is. "Flag" is NOT "option".

The **"options"** = **"opciones"** in the monero-wallet-cli are:
--deamon-address arg
--deamon-host arg 
--password arg...

You should use the **"flag"** = **"marcador"**:
--deamon-address arg, when...
--deamon-address arg, when...
--password arg, when...

These are two different arguments even if "flag" means "option". One indicates what are your options, the other indicates what you should use(**flag**), or do the direct signal in a case of using the OPTION.

We have to differentiate them because in the context is used differently and mean differently, I gave you the [wikipedia "flag" definition in spanish](https://es.wikipedia.org/wiki/Flag) so you could read it in spanish because it clearly means something different. Not sure what @BlackLotus64 had in his mind when using the word "bandera" (that could also work), but he was in the context, and you aren't.

lh1008:
I'll give you another example:

What is `--deamon-address arg` command different than an option?
A flag, why?
Because is used in information processing inside the CLI. So `--deamon-address arg` is an option inside the CLI that gives a signal inside the CLI, and what is called the signal? A flag, not an option. We can keep on arguing what it should be. Just take your time and read the text and ask youruself why was the word "flag" and not "option" used, it should mean something.

BlackLotus64:
I think it doesn't really matter, and yes @lh1008 the page talks more about internal structure of processors and states, but in the end it's the same as flags or options here, they just change values in the daemon and are indicators of that change.

We can ask for a programmer point of view for this, because in my university even professors said _bandera_ in microcontrollers programming, even if we just knew that they were some kind of state/option/indicator.

But in my opinion, _option_ is a bit more, um, let's say, for-all-people, when flag is more technical. In English it can also be _marker_.

BlackLotus64:
Maybe @el00ruobuob ? Because I'm almost new here I don't know a lot of people, but for now let's just stay with this PR, we can discuss later the translations to another language like Spanish, don't you think @lh1008 ?

lh1008:
I just needed to make it clear. I'm always a detail person, this is not a personal engagement with anyone sorry guys.

PR #881, opened 2018-09-09
Wallet RPC documentation correction

woodser:
The 'getbalance' RPC call has a field 'per_subaddress' which is an array but the documented response shows an object.  This PR adds the arrays symbol.

Not sure what the correct procedure is to update the RPC documentation so please let me know if it should be done differently.  I expect other minor corrections.


el00ruobuob:
There is a PR #854 which contains a full refresh of the wallet RPC guide. Would you mind reviewing it instead? I don't remember if I correct this one.

woodser:
Close based on PR #854

PR #892, opened 2020-04-02
workflows: add site CI

selsta:
https://github.com/selsta/monero-site/actions/runs/69283929

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,22 @@
+name: gh-actions/site
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        ruby: [ '2.4', '2.5', '2.6', '2.7' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,22 @@
+name: gh-actions/site
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        ruby: [ '2.4', '2.5', '2.6', '2.7' ]
```
erciccione, line 10: Wouldn't be fine to run tests only on `2.6`? We don't need to test it on multiple versions IMO

selsta: I don’t see why we shouldn’t. Maybe we can catch bugs this way as I don’t think everyone uses the same ruby version. I takes equally long and the computing power is free anyway.

erciccione: Fair enough. If it takes the same time, no problem.

Issue #924, opened 2020-04-12
Change emoji icons to something more consistent


erciccione:
*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/1075*

---
Emojis look different on each device and some can be ugly (for example the check on Chrome appears grey). Right now we use emojis for the developer guides page, FAQ and roadmap.

The options could be:

- Change emojis
- Use minimal black and white icons (https://icons.getbootstrap.com was pointed out as a possibility)
- Use pics (option that i discarded initially to not make the page heavier)
- https://primer.style/octicons/ (can be integrated with jekyll)

Voice your opinion.

alreadyburnt:
So here's 


I think we should use SVGs from [openmoji.org](https://openmoji.org/). They're nice and can be downloaded separately too. Each comes with a black and white version, although I think pages like the roadmap look better with the colored ones. I think we should choose only one source for consistency (eg. not a little from Bootstrap and a little from OpenMoji), but the roadmap page + black and white bootstrap icons looks ugly and boring. Here's how it would look with openmojis.

![screen](https://user-images.githubusercontent.com/81592644/113633392-2ddd1f80-966d-11eb-84b1-48480462956a.jpg)

alreadyburnt:
So I think this is the complete list of what needs to be replaced.
`grep -rIE '"\\0.+"' monero-site/`

> monero-site/css/custom.css:  content: "\02714";
> monero-site/css/custom.css:  content: "\01F6A7";
> monero-site/css/custom.css:  content: "\01F9ED";
> monero-site/css/custom.css:  content: "\01F4C4";
> monero-site/css/custom.css:  content: "\01F310";
> monero-site/css/custom.css:  content: "\02714";
> monero-site/css/custom.css:  content: "\01F6A7";
> monero-site/css/custom.css:  content: "\01F9ED";

I suggest using SVGs from [OpenMoji](https://openmoji.org/). They're nice and can be downloaded separately too. Each comes with a black and white version, although I think pages like the roadmap look better with the colored ones. I think we should use only one source for consistency (eg. not a little from bootstrap and a little from openmoji), but the roadmap page + black and white bootstrap icons looks ugly and boring imo. Here's how it would look with openmojis.

[...]

PR #926, opened 2020-04-12
faq: replace link emoji with svg

selsta:
I’ve found it easier to add the few .svg we use manually instead of importing the Jekyll plugin. I didn’t figure out how to add them to the css file with the plugin.


erciccione:
Yes, that's probably the direction we should be going for #924 (use svg logos).

PR #956, opened 2020-04-28
New 404 page

erciccione:
The old javascript-based 404 page is broken and the standard page from nginx shows up in case of a 404 error.

Removed the '404' folder with its content and created a new `404.html` file in the root directory, added some specific css and slighlty edited the 'error' layout.

@binaryfate might be necessary to slighly tweak the nginx .conf file to have this working properly. See https://jekyllrb.com/tutorials/custom-404-page/#hosting-on-nginx-server

Preview:

Screenshot_2020-04-28


SamsungGalaxyPlayer:
I'd like to have better text under the 404 that's Monero-related somehow. Don't hold the PR for this though.

dginovker:
Definitely a beautiful PR, but I agree with @SamsungGalaxyPlayer.


# 404
_Page not found_

Usually we want Monero to break links. Not this time though.

jerrimus:
### 404

Link not found ( perhaps it values privacy )

erciccione:
No problem in holding the PR, this is not urgent. I'm ok changing the text, but i would prefer to have it on a single line:

404  
<'page not found' text>
<'to the homepage' button>

erciccione:
If there are no more ideas, i would suggest to use a slightly edited version of @jerrimus' phrase:

`Page not found ( perhaps it values privacy )`

PR #958, opened 2020-04-29
README: Add GitHub actions badge for master build status

erciccione:
The badge will tell us if master builds correctly or something is broken

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+![gh-actions/site](https://github.com/monero-project/monero-site/workflows/gh-actions/site/badge.svg?branch=master)
+
 ## Table of Contents
 
 - [Introduction](#introduction)


README.md
```diff
@@ -1,3 +1,5 @@
+![gh-actions/site](https://github.com/monero-project/monero-site/workflows/gh-actions/site/badge.svg)
```
selsta, line 1:
```suggestion
![gh-actions/site](https://github.com/monero-project/monero-site/workflows/gh-actions/site/badge.svg?branch=master)
```

erciccione: If a branch is not specified it uses the default branch selected on GitHub, which in our case is master, that's why i didn't specify it before. Changing it anyway.

selsta: ah true, yea guess then it isn’t necessary

PR #961, opened 2020-05-01
Remove translations of developer guides

### LINKED ISSUE #921: Remove possibility to translate developer guides
*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/1074*

---
The structure of the user and dev guides needs to be changed. I stated in past (for example https://repo.getmonero.org/monero-project/monero-site/-/merge_requests/1243#note_9413), it's extremely hard to work on those files and it's basically impossible to update the guides for all languages. That system is simply broken.

I already fixed the roadmap and i plan to use a similar approach to fix the user guides, but for the dev guides the situation is different. Those guides are very technical and i don't see the benefit of having them translated, i think would be much better to simply have them in English.

Almost all guides are untranslated for all languages, and the few which are translated are outdated and very hard to update because of the reasons i wrote above (and even if we did manage to update them, we would have guides partly translated and partly in English).

So, i propose to remove the possibility to translate the developer guides and only leave the english versions, if there are no comments or objections i will just go ahead and work on this in the next days.

erciccione:
and remove the possibility to translate them. The 2 dev guides ('wallet RPC' and 'Daemon RPC') are now available only in English

Closes #921


selsta:
```
Conflicting files
_i18n/en/resources/developer-guides/wallet-rpc.md
```

erciccione:
Rebased

PR #966, opened 2020-05-04
Simplify versioning system for user guides

erciccione:
This new approach doesn't require version numbers and make easier for contributors to understand the status of the document. There are only 3 states:

- Outdated: The page is outdated and might not work as expected
- Untranslated: The page needs to be translated
- Translation outdated: The page is translated, but the original (English) document has been updated

To mark a guide as outdated, we change the boolean of the front matter entry `outdated` in `/resources/user-guides/GUIDE.`
If 'False', the guide is updated. If 'True' all versions of the guide (English included) will show a warning at the bottom of the page.

The other 2 states are related to the status of the translated pages and we control them from the language-specific
user guides: `/_i18n/LANG/resources/user-guides/GUIDE`. At the top of the page a snippet will include 2 parameters:

- `translated`: "yes" if the page is translated, "no" if it's not.
- `translationOutdated`: "yes" if the translation is outdated, "no" if it's not

This new system aims to be simpler than the precedent, avoiding to compare versioning numbers and using a higher level system instead (yes, no, True, False). I also removed the middle way status 'only minor changes', because if there are only minor changes that don't affect the usability of the guide, we don't need to point it out.

The old system was complex and people didn't use it. These changes aim to make things easier for translators and other contributors.

Notes:

- Snippet `{% include disclaimer.html translated="no" translationOutdated="no" %}` left in english files, so that can easily copied with the rest of the page.
- the default is `outdated: False`. I will check all guides to see if they are updated in another PR.
- Checked if every guide of each language (excluded English) was translated or not and adapted the snippet accordingly.


erciccione:
@00-matt I just rebased and retested the PR but i couldn't see any problem. Also your screenshot show the top of the homepage, which i havent's changed in this PR. What are you referring to?

00-matt:
All of the pages ended up like this after applying the PR. Will retest soon

Issue #965, opened 2020-05-03
Monero Betting


dsmith46v:
#### What's the name of your business?

Monero Betting (Darknet)

#### What's the URL of the website?

http://moneroeetsad6quaq4weauz5ndsx5jf6y3z5zwtckhudciapom26byid.onion/

#### Write a brief description of your business/service (max 50 characters)

Very simple Monero betting game!

#### Enter the URL where a Monero mention can be found

Index page
http://moneroeetsad6quaq4weauz5ndsx5jf6y3z5zwtckhudciapom26byid.onion/

#### How do you accept Monero? *(choose only one option)*

- [ ] Monero integrations *(github.com/monero-integrations)*
- [ ] Globee
- X Native wallet
- [ ] Other payment gateaway (Coinpayments, Morphtoken, etc. *Please specify*)

#### How would you categorize your business? *(choose only one option)*

- [ ] Exchange
- [ ] Block Explorer
- [ ] Payment Gateway
- [ ] Library and Helper
- [ ] Tool
- [ ] Service
- [ ] Goods
- X Entertainment

#### Comment

Secure, private, untraceable betting.
Very simple betting game based on Monero transaction hash.

erciccione:
Hi.

We never listed merchants on onion services. Personally i have no problem to add tor-only websites, as long as they don't offer illegal services/goods. Let's wait for more opinions.

00-matt:
We could have an icon with a tooltip next to all onion websites explaining that you need to use the Tor browser to access them.

erciccione:
As we say on the disclaimer in the merchants page:

> Please note: these links are being provided as a convenience and for informational purposes only; they do not constitute an endorsement by the Monero community of any products, services or opinions of the corporations or organizations or individuals listed. **The Monero community bears no responsibility for the accuracy, legality, or content of these external sites**. Contact the external site for answers to questions regarding its content. As always, caveat emptor ('buyer beware'); 

So, i don't think it should be a problem to list a website hosted on an onion address.

PR #984, opened 2020-05-17
Merchants: Add onion logo when merchant is on onion service

erciccione:
onion

As suggested by @00-matt in https://github.com/monero-project/monero-site/issues/965#issuecomment-623962937


erciccione:
Would b probably better to use a .svg image, since they autoscale and are very light. If i cannot find one for our purposes i will just shrink it as you suggest.

erciccione:
Added svg image (taken [directly from Tor](https://media.torproject.org/image/Onion%20Icon/)) and added simple tooltip using "title".

PR #992, opened 2020-05-22
Downloads: set 'common issues' back to 'noissues' (prep for 0.16)

erciccione:


erciccione:
This should be merged along with #1000

binaryFate:
Makes sense, no brainer.

PR #1004, opened 2020-05-24
Developer Guides: Remove unnecessary curl parameters

00-matt:
Remove `-X POST` from calls to curl. It doesn't do anything good and now the commands are shorter.

[Article by curl author warning against the unnecessary use of -X](https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/)

### DIFF
--- a/resources/developer-guides/daemon-rpc.md
+++ b/resources/developer-guides/daemon-rpc.md
@@ -88,7 +88,7 @@ METHOD='get_block_header_by_height'
 ALIAS='getblockheaderbyheight'
 PARAMS='{"height":912345}'
 curl \
-    -X POST http://$IP:$PORT/json_rpc \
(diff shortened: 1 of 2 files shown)


resources/developer-guides/daemon-rpc.md
```diff
@@ -2181,7 +2181,7 @@ Outputs:
 Example:
 
 ```
-$ curl -X POST http://127.0.0.1:18081/get_limit -H 'Content-Type: application/json'
+$ curl http://127.0.0.1:18081/get_limit -H 'Content-Type: application/json'
```
00-matt, line 2184: The `-X POST` was not unnecessary here as there is no `-d` flag, however I don't think this should be a POST anyway since we just get data. (I tested with GET and POST, you get the same response)

PR #1009, opened 2020-05-25
User guides: specify 'Importing blockchain' is for Windows

### LINKED ISSUE #908: Change title of *Importing Blockchain* user guide
*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/1041*

---
Change title of user guide *Importing Blockchain* to *Importing the Blockchain to Monero GUI wallet (Windows)*.

Ref https://repo.getmonero.org/monero-project/monero-site/merge_requests/1173#note_8479

erciccione:
Closes #908 

I didn't specify *for the gui wallet* because works also for the CLI.

This guide could be updated. The GUI offers a native way to import a blockchain now.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -489,7 +489,7 @@ user-guides:
   wallets: Wallets
   offline-backup: How to make an offline backup
   vps-node: How to run a node on VPS
-  import-blockchain: Importing the Monero blockchain
+  import-blockchain: Importing the Monero blockchain (Windows)
   monero-tools: Monero Tools
   purchasing-storing: Securely purchasing and storing Monero
   verify-allos: Verify binaries on Linux, Mac, or Windows command line (advanced)


00-matt:
I think that the only Windows specific parts are the drive letters and opening CMD. Would be useful to users on any operating system imo

erciccione:
I see your point but the guide is still specifically aimed to Windows users. If we feel it's not a big deal, i would be fine with keeping it as it is.

00-matt:
Maybe the page could be deleted and the link on the downloads page be replaced with "just use `monero-blockchain-import --input-file blockchain.raw`"? Along with a warning that it will probably be slower than syncing through the daemon.

erciccione:
I feel that it would make the guide useless. Maybe we could add a FAQ entry like "How do i import a blockchain?", add your answer, a disclaimer and the link to the user guide,

00-matt:
That sounds like a nice solution to me.

ndorf:
I tried to improve the document somewhat [here](https://github.com/monero-project/monero-site/pull/1065), but I think it might be better to simply remove it and replace it with a brief FAQ entry, as suggested here.

One way or another, IMHO, new users should be discouraged from using this function altogether. In earlier times, maybe the P2P network was slower and/or less reliable than a direct download, but that has not been the case in years. For some reason, though, new users continue to incorrectly assume that downloading and importing will be faster. Then they end up using the dangerous unverified import option, which the current guide openly encourages (the main impetus for my PR).

It could still be useful sometimes for advanced users and developers, but those don't need multi-step instructions on how to open a CMD window. A brief FAQ entry with the basic command and (IMO) strongly-worded warning about the dangerous should suffice.

erciccione:
> new users should be discouraged from using this function altogether.

Thanks for the PR. Consider that users are warned about importing the blockchain being slower than the normal sync in the download section. Being more clear doesn't do any arm tho.

>  remove it and replace it with a brief FAQ entry,

I think that guide should stay. Users might have their own reason to prefer downloading the blockchain the slow way. Let's just warn them about the downside and then let them do whatever they want.

Anyway, I'm closing this PR and #908, since as said, even if the guide is aimed to Windows users will work for everybody else.

PR #1010, opened 2020-05-25
update genfund donation to subaddress -- still same wallet

### LINKED ISSUE #985: Replace integrated address with subaddress for Donations
There are still several 4... addresses, for example on https://web.getmonero.org/get-started/contributing/ should we replace with 8... addresses?

In my understanding Subaddresses are principally preferable before 4...main-addresses. So is it possible to replace it, also here on Github and other places like https://revuo-monero.com/issue-60.html

We would want to propagate the use of 8... addresses, right?

binaryFate:
Use a subaddress for donations instead of the 4.. address.

It is the same wallet, so the publicly known view key remains valid.

Fixes #985

### DIFF
--- a/_i18n/ar/resources/moneropedia/address.md
+++ b/_i18n/ar/resources/moneropedia/address.md
@@ -6,7 +6,7 @@ summary: "either an alias, such as donate.getmonero.org, or a set of 95 characte
 {% include untranslated.html %}
 ### The Basics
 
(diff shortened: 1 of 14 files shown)


binaryFate:
Same update for CCS repo: https://repo.getmonero.org/monero-project/ccs-front/-/merge_requests/30

PR #1012, opened 2020-05-26
Fixed Bug Referenced in #1007

### LINKED ISSUE #1007: Clicking homepage video to pause will not pause
When I click "Play" on the video on the homepage, it will play.

However, if I click the body of the video in order to pause it, it will not pause. It only pauses when I click the pause button.

I could see how this could be determined as "expected behavior", but I thought I would bring it someone's attention as it surprised me.

I am using Chrome Version 81 on Windows 10.

I reproduced using Chrome, Version 81 on Windows 10 on browserstack.

Would this project be open to using an open source media player for videos? I've found the HTML5 video element to be quite buggy in it's implementation throughout browsers and that might be a more simplistic solution in the long run.

Samuel-Pedraza:
I added some JS in the bottom of _includes/footer.html, I wasn't sure where to add it.

I can refactor if needed.

Just to update - I verified this works on Chrome, Internet Explorer, Edge, and Firefox on Windows 10

Fixed #1007


Samuel-Pedraza:
Thanks, I'll try and figure out another way to handle it.

Samuel-Pedraza:
Closing pull request as it seems to no longer be needed due to most recent Chrome update.

PR #1032, opened 2020-06-07
home: remove 'reasonably' from header

### LINKED ISSUE #971: Remove "reasonably" from "reasonably private" on the homepage
Initial MR: https://repo.getmonero.org/monero-project/monero-site/-/merge_requests/1136

Related discussion: https://repo.getmonero.org/monero-project/monero-site/-/merge_requests/1219

Related discussion: https://www.reddit.com/r/Monero/comments/f2m7la/a_reasonably_private_digital_currency_why_say/

While I agree that being careful about wording is prudent, I think this wording causes more confusion than anything else. We should point people to other, more detailed resources like the MRL papers and Breaking Monero if they are interested in understanding this further.

We can add some text to the downloads page description saying something along the lines of the following:

---

On this page you can find and download the latest version available of the Monero software, as well as hardware, light and mobile wallets. Monero is experimental software. **While it provides far greater privacy out of the box than other tools, you should understand the limitations if you are concerned about your privacy.** Using Monero without changing your habits, as with using Tor, isn't sufficient for all use-cases.

erciccione:
closes #971

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -52,7 +52,7 @@ index:
   page_title: "Monero - secure, private, untraceable"
 
 home:
-  heading2: A Reasonably Private Digital Currency
+  heading2: A Private Digital Currency
   monero_is_cash: Monero is cash for a connected world. It’s fast, private, and secure. With Monero, you are your own bank. You can spend safely, knowing that others cannot see your balances or track your activity.
   get_started: Get Started
   why_monero_is_different: Why Monero is different


dEBRUYNE-1:
Should we remove the 'A' too? It arguably reads more fluent when it is left up.

erciccione:
For some reason i assumed that was the case. I can put the "A" back tomorrow.

erciccione:
Done

erciccione:
Could somebody approve this so we can merge?

PR #1046, opened 2020-06-15
home: fix video not playing if localized file doesn't exist

erciccione:


index.html
```diff
@@ -11,7 +11,11 @@
                         <div class="monero-video">
                             <!--iframe width="560" height="315" src="https://www.youtube.com/embed/TZi9xx6aiuY" frameborder="0" allowfullscreen></iframe-->
                             <video controls poster="/img/monero-community.png" preload="metadata">
-                                <source src={% if site.lang == 'en' %}"/media/Monero_Promo.m4v"{% else %}"/media/{{ site.lang }}/Monero_Promo.m4v"{% endif %}>
+                                {% if site.lang != 'pt-br' and site.lang != 'ru' %}
```
selsta, line 14: Would it make more sense to have site.has_video .yml entry instead of hardcoding it here? Would that be possible or overkill?

erciccione: I thought a fair bit about this. It's not optimal, but i do think that anything else would be overkill at the moment. We have the videos translated in these two languages and i don't think we will have more voice-over translations soon. If that will happen, we can just think of a better way to handle it.

selsta:
please rebase

index.html
selsta: sounds ok in this case

erciccione:
Rebased

PR #1027, opened 2020-06-05
Blog: Use only one 'Blog' menu and add category selector

erciccione:
- Added separator between posts
- remove all submenus and use only 'blog'
- Add category selector for desktop and mobile

Resolves  #1001

Preview:
blog-preview

PR #1002, opened 2020-05-24
Meta tags: Change opengraph logo and add possiblity to use custom one

erciccione:
The old logo showing on social medias is too large and end up being trimmed:

Twitter:
twitter

Reddit:
reddit

Replaced it with the simple Monero Logo. If a logo is specified using 'image:' in the front matter, jekyll will show this one instead. This also allow us to easily make specific meta-images for each blog post. I will add the possibility to add the image in the body of the post in a subsequent PR.

PR #1051, opened 2020-06-22
Improve RSS experience (install 'jekyll-feed' plugin)

### LINKED ISSUE #916: Improve RSS experience
*Issue originally created by @sanecito*

*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/1066*

---
Problem: Currently the RSS feed isn't in a location where it's likely to be seen and it isn't detected by Akregator when using the getmonero.org URL to add a feed. Improving the RSS feed experience drives RSS adoption and helps increase readership of blog posts.

Solution: 

1. Put RSS feed somewhere that isn't the very bottom of the website
2. Look into what's required for RSS readers like KDE's Akregator to detect RSS feeds on getmonero.org

erciccione:
Based on #1027 (review only 283479c)

- Use ['jekyll-feed' plugin](https://github.com/jekyll/jekyll-feed) to manage RSS feed and remove existing feed.xml file
- Add feed image in main blog tab ('All posts')

The 'jekyll-feed' plugin makes possible for aggregators to automatically detect the existence of a feed (feed_meta tag in header) and creates a 'feed.atom' file containing the last 20 blog posts and locates it in the root directory. The structure is compatible with our current system, including the possibility to show a picture if it's included in the post (#1002).

The plugin is one of the most used jekyll plugins so it will likely be maintained for a long time (it's suggested on the official jekyll website).

Closes #916


erciccione:
@sanecito could you test this PR (you can use the netlify preview above ^) to make sure your RSS reader correctly detects the feed?

sanecito:
@erciccione Can confirm it works with KDE's Akregator's feed detection:

image

Technically is stuck downloading, but that might be due to the preview nature of it?

erciccione:
> Technically is stuck downloading, but that might be due to the preview nature of it?

Maybe. I don't have much experience with feed aggregators so i'm not really sure about how the autodetection works. For what it worth, i can download the atom file with no problems from netlify's preview.

erciccione:
I did some tests using Liferea and the autodetection fails there as well. The reader seems to be looking for the feed on getmonero.org and gives a 404 error when it doesn't find it. Maybe akgregator goes in a loop instead of prompting the error right away.

Autodetection seems to be working fine when testing locally, so i would say it's a problem related to the preview.

sanecito:
So should be good to merge then, right?

erciccione:
Yes, it should be.

_layouts/blog_by_tag.html
```diff
@@ -9,21 +9,55 @@
...
-		                </div>
-                    </div>
-                    
+                          <div class="page-title">
+                          {% if tag.name != "Dev Diaries" %}
+                            <h2 class="inline"><span class="kicks">{{ tag.name }}</span></h2>
+                          {% else %}
+                            <h2 class="inline"><span class="kicks">{% t blog.meetinglogs %}</span></h2>
+                          {% endif %}
+		                      </div>
```
selsta, line 56: mixes spaces and tabs

erciccione: I still have to figure this out. The website is mixed. Some documents use spaces and some use tabs. Some pages are even mixed and they use both tabs and spaces. This file is a good example of my confusion.
I often refrain to change indentation to avoid useless diffs, but i think we should decide for either using tabs or spaces and adapt the files we edit accordingly.

In this case i might have just made confusion, because i see i changed indentation. I will just use tabs here, but the problem remains.

selsta: I would not change existing code, just try to not mix it when adding new stuff

erciccione: My point is that "not change existing code" and "not mix it when adding new stuff" is not possible. Whatever i do (without reindenting everything), it will be mixed. The part i added (the top) uses tab, but the bottom uses mixed tabs and spaces.

My additions at the bottom are mixed because whatever i choose to do i have to change indentation of other parts, which i wanted to avoid. Do you suggest to reindent all the parts i touch?

selsta: Not really that important but the lines beneath and above are all spaces, so I would just change this line to all spaces.

_layouts/blog_by_tag.html
```diff
@@ -9,21 +9,55 @@
...
+                        </div>
                        {% if site.tags[tag.slug] %}
-				{% for post in site.tags[tag.slug] %}
-                  <div class="post-tag">
+			                 	 {% for post in site.tags[tag.slug] %}
```
selsta, line 59: Here too, beneath and above are all spaces so this should be all spaces too.

_layouts/blog_by_tag.html
erciccione: Alright. i'll do that.

erciccione: Oddly enough, the indentation looks good on VS code, but looks off as shown here with another editor. i will have to check some settings i guess.

selsta: Github probably uses a different visual tab width.

_layouts/blog_by_tag.html
erciccione: Should be fixed now. Please let me know if anything else looks off.

PR #1066, opened 2020-07-10
blog: add post for GUI release 0.16.0.2

selsta:


dEBRUYNE-1:
Even though the fix for the Ledger bug (i.e. the bug that inhibits Ledger Monero users from properly sending transactions with multiple inputs) is basically present in the core repository, I'd recommend, for visibility purposes, to put it in this blog post as well.

erciccione:
Sounds like we should also mention the problem in the "common issues" section in the download page.

dEBRUYNE-1:
@erciccione - Not sure if we should, as it is now basically fixed in both the CLI and GUI release.

PR #1086, opened 2020-07-17
workflows: improve performance of github CI

erciccione:
- Bump actions/checkout to v2
- Limit number of posts to be built to 10

Build time reduced from about 6 minutes to 2.20 minutes (rough average)


erciccione:
@00-matt See https://github.com/monero-project/monero-site/pull/892#discussion_r407413181. Personally i'm ok with living it as it is.

PR #1109, opened 2020-07-30
Add whitepaper_annotated.pdf and whitepaper_review.pdf

erciccione:
and link to them from the 'Research Lab' page. These files were initially hosted on downloads.getmonero.org.


SarangNoether:
Note that the documents still exist in `monero-downloads` to avoid breaking any existing links to them. If updated here, they should also be updated there.

PR #1111, opened 2020-07-31
Add whitepaper reviews

SarangNoether:
Adds whitepaper reviews from the `monero-downloads` repository and updates links. The documents still exist in `monero-downloads` to avoid redirects.


SarangNoether:
Covered by #1109 already!

PR #1118, opened 2020-08-04
head.html: show monero logo in og:image if page.image is an SVG file

erciccione:
SVGs are not supported by opengraph.


selsta (approved):
Untested, code looks ok

PR #1131, opened 2020-08-12
Remove 'Libraries and helpers' from the merchants page and replace it with 'Web Hosting' + Add ''Libraries and helpers" to the dev guides

### LINKED ISSUE #974: Move Libraries and Helpers away from merchants page
From IRC:

```
<asymptotically> should we move Libraries and Helpers out of the merchants page?
<@ErCiccione[m]> asymptotically: why?
<asymptotically> well they aren't really merchants, they are tools for developers that want to integrate with monero
<SerHack> +1 for the idea
<@ErCiccione[m]> asymptotically: Sure, if we find another place i'm ok with that
<@ErCiccione[m]> maybe the page of the guides
<@ErCiccione[m]> i recently added more resources there
<asymptotically> we could have it as another page under the community tab and also link to it from the guides page?
<@ErCiccione[m]> I wouldn't add another tab, the menus are already very clogged imo
<@ErCiccione[m]> we could have simply another page and we link to it from the 'guides' page and wherever we want
```

erciccione:
These two changes have to be made together because, with the current structure, if we remove a section from the merchants page we have to replace it with something else.

- **Moved 'Libraries and helpers' from the merchants page to the developer guides**. Most of the listed repositories were obsolete. I removed almost all of them and replaced with maintained repos (if anybody have more repos to suggest, please let me know).
- **Added section 'Web Hosting' to the merchants page**. We had a big amount of merchants offering web hosting in the 'services' section. Since to not mess up the selector at the top we need to keep the same amount of categories in the merchants page, i thought was a good idea to create a section dedicated to web hosting.

Closes #974

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -137,35 +137,49 @@
       url: https://swirlpay.io/
     - name: Vigla
       url: https://vigla.biz/
-- category: Libraries and Helpers
-  id: libraries
+- category: Web hosting
+  id: hosting
   merchants:
-    - name: Monero Integrations (eCommerce plugins)
-      url: http://monerointegrations.com/
-    - name: monero-nodejs (Node.js)
-      url: https://github.com/PsychicCat/monero-nodejs
-    - name: MoneroApi.Net (.NET)
-      url: https://github.com/Jojatekok/MoneroApi.Net
-    - name: moneronjs (NodeJS)
-      url: https://github.com/netmonk/moneronjs
-    - name: Monero Payments Library (PHP)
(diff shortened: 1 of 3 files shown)


erciccione:
I don't know if it worth it to be honest. Would become a simple list of urls. Unless you had something different in mind?

00-matt:
Yes just a list of names and URLs in categories like how the merchants
are now. Not really worth it for the 10 or so entries in there at the
moment.

erciccione:
Rebased.

PR #1136, opened 2020-08-16
Adopt versioning system used in the User Guides to keep track of the translation status of the Moneropedia entries

erciccione:
- Removed the 'untranslated.html' snippet (_includes/untranslated.html). Since now we are using 'disclaimer.html' everywhere
- Replaced the old snippet with the new disclaimer (introduced with #966) in all Moneropedia entries
- Updated instructions in the README

Add versioning snippet to all 'randomx' Moneropedia entries (except en)


erciccione:
Translators are currently working on these files. It's necessary to review this PR so can be merged, otherwise translators will be working on outdated files.

00-matt (commented):
Seems like some untranslated entries (like the [Dutch translation of Block](https://deploy-preview-1136--barolo-time-757cf9.netlify.app/nl/resources/moneropedia/block.html)) are missing the disclaimer. Will check some of the other pages too to see if it needs adding to any other pages.

erciccione:
@00-matt I didn't add the disclaimer to any dutch entry. I don't remember if there was a reason behind that choice or i just missed it. Anyway. I'm double checking all languages and adding the snippet where it's missing.

erciccione:
So, there were some conflicts because some stuff got merged in the meantime, not sure why github's UI wasn't showing it. Also, i only edited the existing snippet when i opened the PR. Now i also added the new snippet where there wasn't any.

There are some open PRs which remove some moneropedia entries. I didn't edit these entries to avoid conflicts (they will be removed anyway).

Also, for some reason pt-br entry was in Italian. No idea of why that happened but i replaced the entry with the english version.

I took advantage of the fact that i was already touching the PR to add the snippet in the also in the english files, this will make things easier when adding new languages (the snipper will be already there). Let's keep in mind that, as i stated many times, the structure of the moneropedia is completely messed up and was left outside the localization plugin, so the entire section needs to be refactored anyway.

I also reorganized the commits, squashing the two original ones in only one.

00-matt (approved):

PR #1138, opened 2020-08-18
Add 'Circled Information Source' symbol to Moneropedia links

erciccione:
Screenshot from 2020-11-04 14-59-45

Tested on Firefox and Chromium. Please test it on more browsers.

I was thinking of removing the dotted bottom to the internal Moneropedia links. Mostly because i don't like that in general and i thought the link to the Moneropedia entry would look nicer only in bold + the circled 'i' logo. I decided to keep the dotted bottom here only for consistency with the other internal links, but if people agree it would look nicer without, i'll take it out with pleasure :)

### DIFF
--- a/_plugins/moneropedia.rb
+++ b/_plugins/moneropedia.rb
@@ -120,14 +120,14 @@ module Jekyll
           if lang != "en"
             @@moneropedia[lang].each do |entry|
               entry[:terms].each do |term|
-                content = content.gsub(/(\@#{term})((?=-#{lookahead})|(?![\w-]))/i, '<a data-tooltip="' + entry[:summary] + '" href="/' + lang + '/resources/moneropedia/' + entry[:file] + '.html" >' + term.gsub('-',' ') + '</a>')
+                content = content.gsub(/(\@#{term})((?=-#{lookahead})|(?![\w-]))/i, '<a class="info-tooltip" data-tooltip="' + entry[:summary] + '" href="/' + lang + '/resources/moneropedia/' + entry[:file] + '.html" >' + term.gsub('-',' ') + '</a>')
               end
             end
           end
           # For English and as a default (in case an entry has been forgotten in a non-English moneropedia folder)
           @@moneropedia["en"].each do |entry|
             entry[:terms].each do |term|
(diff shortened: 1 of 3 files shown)


erciccione:
Converted to draft because i'm currently working on changing the way external links are displayed and i want to make consistent styling choices.

erciccione:
Showing external links automatically on GetMonero it's trickier than i anticipated, Let's merge this in the meantime, we can adapt the styling later if necessary.

erciccione:
Showing external links automatically on GetMonero it's trickier than i anticipated, Let's merge this in the meantime, we can adapt the styling later if necessary.

erciccione:
Rebased to update the preview. This still needs to be reviewed.

selsta:
Screenshot 2020-10-18 at 23 14 59

Looks like this in Safari

erciccione:
@selsta could you test now?

selsta:
same hmm

erciccione:
I have no clue. Are we sure it's not a local issue? Maybe fonts or something else?
Safari should support Unicode entities.

00-matt:
[Also broken in Chrome on macOS](https://i.imgur.com/uDGUgsg.png). Just a guess, but maybe you could use `content: "\1F6C8"` in the CSS instead??

erciccione:
> maybe you could use content: "\1F6C8" in the CSS instead?

That's what i did before force pushing the last change. Using `\1F6C8` should give less compatibility problems reguardless, but i thought it worthed a try.

erciccione:
The icon seems to work correctly in my local machine in Firefox and chromium, but doesn't work on my ubuntu VM in chromium, chrome and firefox. I'm considering switching to just use an SVG. Making this PR a draft until i figure it out.

erciccione:
The problem was not related to browsers but to fonts. Open sans doesn't include that unicode character so you either have it installed locally or it won't show up. The only reliable solution is to use an svg image, so i uploaded one.

To add some stylistic diversity i removed the dotted line under these links. Now there is the usual continuous line used for links on hover. I updated the screenshot.

This is now ready to be reviewed.

00-matt (approved):

PR #1144, opened 2020-08-19
ci: Fix hash checking workflow

00-matt:
* Download bF's key from GitHub instead of from unreliable keyserver
* Resolve redirects manually instead of checking server

[Example run](https://github.com/00-matt/monero-site/runs/1004401847)


00-matt:
@binaryFate requires the release tarballs to be available on downloads.getmonero.org before you add the hashes to monero-site, but it does not require the redirects to be updated.

erciccione (approved):

PR #1145, opened 2020-08-20
Add icon to external links

erciccione:
This PR and #1138 aim to give a better UX by clarifying to the user what type of link they are about to click. In this specific case i think it's important to tell the final user they are going to be redirected outside getmonero.

### DIFF
--- a/community/hangouts/index.md
+++ b/community/hangouts/index.md
@@ -10,19 +10,19 @@ permalink: /community/hangouts/index.html
     <section class="hangouts-social container">   
         <ul class="row center-xs">
             <li>
-                <a href="https://twitter.com/monero" target="_blank" rel="noreferrer noopener"><div class="social-icon twitter"></div></a>
+                <a class="ext-noicon" href="https://twitter.com/monero" target="_blank" rel="noreferrer noopener"><div class="social-icon twitter"></div></a>
             </li>
             <li>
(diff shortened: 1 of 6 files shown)


community/hangouts/index.md
```diff
@@ -10,19 +10,19 @@ permalink: /community/hangouts/index.html
     <section class="hangouts-social container">   
         <ul class="row center-xs">
             <li>
-                <a href="https://twitter.com/monero" target="_blank" rel="noreferrer noopener"><div class="social-icon twitter"></div></a>
+                <a id="ext-noicon" href="https://twitter.com/monero" target="_blank" rel="noreferrer noopener"><div class="social-icon twitter"></div></a>
```
00-matt, line 13: Having multiple elements with the same `id` is invalid HTML.

```suggestion
                <a class="ext-noicon" href="https://twitter.com/monero" target="_blank" rel="noreferrer noopener"><div class="social-icon twitter"></div></a>
```

00-matt (changes requested):
A bit nitpicky because modern browsers accept any old garbage, but I think it's better to try and be correct :)

community/hangouts/index.md
erciccione: Fixed. Thanks.

00-matt (approved):

PR #1146, opened 2020-08-20
Merchants: Fix merchants not using https

erciccione:


_data/merchants.yml
```diff
@@ -42,7 +42,7 @@
     - name: EnterChange
       url: https://enter-change.com
     - name: ExchangeD.I2P - Darknet exchange (I2P)
-      url: http://www.exchanged.i2p/en/home
+      url: https://www.exchanged.i2p/en/home
```
SarangNoether, line 45: Unable to verify this link.

erciccione: It's an i2p address. I just removed the `s` from `https` because apparently they don't need it.

SarangNoether: I know. But I wasn't able to directly test it and therefore can't attest to its correctness.

PR #1163, opened 2020-08-30
FAQ: add 'Can I manually import the blockchain?'

### LINKED ISSUE #1024: Add FAQ "Can i import the blockchain manually?" or something like that
see https://github.com/monero-project/monero-site/pull/1009#issuecomment-637392812

erciccione:
Closes #1024

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -378,6 +378,9 @@ faq:
   aantivirus1: The problem is being discussed and solutions are being elaborated. In the meantime, if you get a warning from your antivirus, make sure the software you downloaded is legitimate (see the guides linked below), then add an exception for it in your antivirus, so that it won't get removed or blocked. If you need assistance, feel free to contact the community.
   qmoneromeaning: "Why is Monero called 'Monero'?"
   amoneromeaning: "Monero is an Esperanto word which means 'coin'. Initially Monero was called 'Bitmonero', which translates to 'Bitcoin' in Esperanto. After the community decided to fork from the original maintainer, 'bit' was dropped in favour of simply 'Monero'."
+  qimporting: Can I manually import the blockchain?
+  aimporting: >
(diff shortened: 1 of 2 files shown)


_i18n/en.yml
```diff
@@ -378,6 +378,9 @@ faq:
...
   qmoneromeaning: "Why is Monero called 'Monero'?"
   amoneromeaning: "Monero is an Esperanto word which means 'coin'. Initially Monero was called 'Bitmonero', which translates to 'Bitcoin' in Esperanto. After the community decided to fork from the original maintainer, 'bit' was dropped in favour of simply 'Monero'."
+  qimporting: Can I manually import the blockchain?
+  aimporting: >
+    Yes, you can, but you probably shouldn't. Importing an external @blockchain is very resource intensive and forces you to trust the entity providing you with the blockchain. It's usually faster to download it the normal way: running a node and letting it syncronize with the other @nodes in the network. If you really need to import an external blockchain, you can download one in <a href="{{ site.baseurl }}/downloads/#blockchain">the 'Downloads' page</a> of this website. Follow the guide below if you are using Windows. If you are a linux user, you can use the tool `monero-blockchain-import`, which is included in the archive when you download the GUI or CLI wallets. Start syncing the imported blockchain with this command: monero-blockchain-import --input-file blockchain.raw.
```
00-matt, line 383: Doesn't seem like markdown works in the language (need to use `<code>` instead of backticks) files, also `s/syncronize/synchronize/:

```suggestion
    Yes, you can, but you probably shouldn't. Importing an external @blockchain is very resource intensive and forces you to trust the entity providing you with the blockchain. It's usually faster to download it the normal way: running a node and letting it synchronize with the other @nodes in the network. If you really need to import an external blockchain, you can download one in <a href="{{ site.baseurl }}/downloads/#blockchain">the 'Downloads' page</a> of this website. Follow the guide below if you are using Windows. If you are a linux user, you can use the tool <code>monero-blockchain-import</code>, which is included in the archive when you download the GUI or CLI wallets. Start syncing the imported blockchain with this command: <code>monero-blockchain-import --input-file blockchain.raw</code>.
```

erciccione: That was on purpose. I realized markdown wouldn't have been rendered while writing the entry, but i preferred to leave the backticks instead of adding more html code, which is confusing for translators. I also didn't want to overcomplicate the structure by leaving the code outside the placeholder. Now while i'm writing i'm tempted to just use <code> and hope translators will not forget it, but if they do it will result in broken html showing. Not sure about this. Opinions?

> s/syncronize/synchronize/

Thanks. i'll fix it tomorrow :)

00-matt: I do think having commands inside code blocks helps with readability, but I don't feel too strongly about it. I wouldn't want to block any new content over a small issue :)

erciccione: I did a couple of tests and i decided to change backticks to double quotes to make the commands more visible. I still prefer to avoid using `<code>`.

00-matt (approved):

Issue #904, opened 2020-04-12
Create a .onion mirror of the site


erciccione:
*Issue originally created by @Electricsheep01*

*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/958*

---
It is quite probable that the FATF is going to start cracking down on privacy focused projects like Monero soon. We should really have .onion mirror(s) of getmonero.org for those who need that extra layer of protection.

erciccione:
@danrmiller Please let us know when there are progresses on this :)

erciccione:
ping @danrmiller

garlicgambit:
Any updates on this issue?

erciccione:
We are finalizing the last details. The onion address will be published soon. (related: https://github.com/monero-project/monero-site/pull/1458)

garlicgambit:
Nice  
  
Are the .onion domains using the single hop/non-anonymous mode?  
Will the CCS also become available as an .onion service?

erciccione:
> Are the .onion domains using the single hop/non-anonymous mode?

I'm not sure, but that would be ok. We don't really need anonymity in this context.

>  Will the CCS also become available as an .onion service?

There are no plans for this currently.

PR #1164, opened 2020-08-30
Make baseurl empty and adapt README, netlify.toml and the sitemap_generator accordingly

### LINKED ISSUE #1162: Duplicate link URL in feed.xml
URL in feed.xml is duplicated as below.

```
<link href="https://www.getmonero.org/https://www.getmonero.org/2020/08/22/triptych.html" rel="alternate" type="text/html" title=" Triptych accepted for publication"/>
```

https://web.getmonero.org/feed.xml

Maybe you need to modify the `baseurl` setting in `_config.xml`?

erciccione:
We enforce a baseurl with a wrong structure (the baseurl [shouldn't be a full URL](https://byparker.com/img/what-is-a-baseurl.jpg)) and that conflicts with jekyll's structure. As a result, most of the plugins, like jekyll-feed don't behave as they should and break. This was also the cause of several other problems we had in past, like the broken sitemap generator and autodetection of the feed.

A `baseurl` is needed for the i18n plugin to work, so we just feed it with an empty one. The result is that all internal links will be stripped of the domain `https://www.getmonero.org', but the structure of the website shouldn't be affected. I made several tests and i don't expect any disruption, unless some nginx configuration gets in the middle for some reason.

This will also make the configuration of the onion mirror easier (#904).

This PR also reverts commit c235ca1db7140e8d6a68ff73c5ebdaa70bfc2004.
Closes #1162

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ If you need support about something related to the website, plese join `#monero-
 Once you have the above list of things, it's typically a good idea to build the website from your local computer to make sure it works before you make any changes. To do this, complete the following steps:
 
 1. Navigate to your local `monero-site` repository.
-2. Serve the website: `bundle exec jekyll serve --baseurl ''`. If you want, you can speedup thi process by loading only the last blog post instead of all of them. Simply add `--limit_posts 1` to the command above. The resulting command will be `bundle exec jekyll serve --limit_posts 1 --baseurl ''`.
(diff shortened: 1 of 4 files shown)


erciccione:
The netlify preview seems to be failing for unrelated reasons.

fluffypony (approved):
Reviewed

PR #1169, opened 2020-09-01
Improve accessibility

erciccione:
I reviewed the entire website to assess its accessibility status. It's in a good condition overall and i haven't found any critical problems. I made some improvements that will improve the UX of visually impaired users. I couldn't implement some changes because they would have required javascript (video controls and tabs scrolling). Would be great if somebody with more experience about accessibility could review this PR, so to give suggestions on how to improve the website further.

- Fixed some interactive elemets which weren't focusable (using 'tabindex')
- Label interactive elements which weren't labelled
- Slightly increase contrast of the grey text in the top navigation bars to reach level AA in the [WCAG](https://www.w3.org/TR/WCAG20/) rating system and increse visibility of the text

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -761,3 +761,8 @@ blog:
 
 tags:
   notags: There are no posts for this tag.
+
+accessibility:
+  arrowup: Go back to the top of the page
+  guiscreen: A screenshot of the Monero GUI wallet. It shows the wallet's balance and a navigation menu on the left, and a form for sending XMR on the right.
+  cliscreen: CLI wallet screenshot
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -10,7 +10,7 @@
                            {% if item.subfolderitems %}
                             {%assign counter=counter | plus: 1%}
                             <label for="drop{{counter}}">{{item.title}}<div class="arrow-down"></div></label>
-                            <input class="burger-checkdropdown" id="drop{{counter}}" type="checkbox">
+                            <input class="burger-checkdropdown" id="drop{{counter}}" type="checkbox" tabindex="0">
                             <div class="dropdown-content">
                                 {% for entry in item.subfolderitems %}
                               <a href="{{site.baseurl}}/{{ entry.url }}">{{ entry.page }}</a>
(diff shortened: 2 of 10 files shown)


_i18n/en.yml
```diff
@@ -763,3 +763,8 @@ blog:
 
 tags:
   notags: There are no posts for this tag.
+
+accessibility:
+  arrowup: "'Back to top' button"
+  guiscreen: GUI wallet screenshot
```
00-matt, line 769: Maybe this could have a bit more detail, something like:

>A screenshot of the Monero GUI wallet. It shows the wallet's balance and a navigation menu on the left, and a form for sending XMR on the right.

It's the best thing to do for images, but I'm not sure how useful it will be when showcasing GUI software (do screenreaders work with the GUI wallet?) and it's another string that will need translating.

_i18n/en.yml
```diff
@@ -763,3 +763,8 @@ blog:
 
 tags:
   notags: There are no posts for this tag.
+
+accessibility:
+  arrowup: "'Back to top' button"
```
00-matt, line 768: A bit more natural when using a screenreader, and makes it more similar to what other sites do when describing links:

```suggestion
  arrowup: "Go back to the top of the page"
```

00-matt (commented):
When navigating with the keyboard only (using the tab key), I can't seem to activate any of the dropdown links at the top of the page (Get Started, Community, and Resources).

erciccione:
> When navigating with the keyboard only (using the tab key), I can't seem to activate any of the dropdown links at the top of the page (Get Started, Community, and Resources).

That's what i meant with *I couldn't implement some changes because they would have required javascript (video controls and tabs scrolling)*. I didn't find a way to open dropdown menus using only tab + return. Seems like Javascript is needed for that to work: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex. I might have missed something though.

_i18n/en.yml
erciccione: Yeah that would be better. I added it last minute, i agree making it mroe verbose makes sense. No problem with adding another string for translators. We can also add explanations of the context is tha string is not self explanatory.

> do screenreaders work with the GUI wallet?

I don't really know.

erciccione: Added both your suggestions.

00-matt:
I think it's weird that pressing tab doesn't go down to the links under the dropdown. They might not be visible but they still exist. I'll have a look at some other sites that have no-JS dropdowns and see how they behave.

erciccione:
Yeah, i agree, but i just made some more tests and i didn't manage to have it work This is the quote from the link i posted that led me to think it's not possible:

> A negative value is useful when you have off-screen content that appears on a specific event. The user won't be able to focus any element with a negative tabindex using the keyboard, but a script can do so by calling the focus() method.

also: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard

00-matt (approved):
:heart:

erciccione:
Rebased

PR #1170, opened 2020-09-01
Blog: add post 'A note on scheduled protocol upgrades'

erciccione:
Posted in place of @dEBRUYNE-1. Posting my review below.

### DIFF
--- a/_posts/2020-09-01-note-scheduled-upgrades.md
+++ b/_posts/2020-09-01-note-scheduled-upgrades.md
@@ -1,0 +1,34 @@
+---
+layout: post
+title: A note on scheduled protocol upgrades
+summary: Elaboration of the reduction in frequency of scheduled protocol upgrades.
+tags: [community]
+author: dEBRUYNE, Justin Ehrenhofer, and community members
+---
+
+Lately, there has been some misconception about the scheduled protocol upgrade policy as well as the original intent of the policy. The goal of this post is to provide further clarity to stakeholders.
+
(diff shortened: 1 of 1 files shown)


_posts/2020-09-01-note-scheduled-upgrades.md
```diff
@@ -0,0 +1,34 @@
+---
+layout: post
+title: A note on scheduled protocol upgrades
+summary: Elaboration of the reduction in frequency of scheduled protocol upgrades.
+tags: [community]
+author: dEBRUYNE & sgp with input and corrections from community members
```
erciccione, line 6: Would be better to have a shorter text here, since this field is not meant to display such long string. I and @dEBRUYNE-1  couldn't come up with something shorter with the same meaning. So leaving it as it is for now, unless somebody has a better solution.

erciccione (commented):
The internal links need to be replaced by liquid links. Doing it now.

erciccione (commented):
The post looks good to me. Would be good to have an approval from both authors @SamsungGalaxyPlayer and @dEBRUYNE-1 (the commit is co-authored by them) and another reviewer. Personally, i think an input from core would be also welcome.

_posts/2020-09-01-note-scheduled-upgrades.md
SamsungGalaxyPlayer: "dEBRUYNE, Justin Ehrenhofer, and community members"

erciccione:
I proposed that change too, but @dEBRUYNE-1 felt it wasn't descriptive enough. If it's fine for him i'll make the change.

SamsungGalaxyPlayer:
I know it makes it longer which is what we're trying to avoid, but I prefer not to be credited as sgp anymore where possible.

dEBRUYNE-1:
Let's go with 'dEBRUYNE, Justin Ehrenhofer, and community members' then. Using 'dEBRUYNE, Justin Ehrenhofer with input and corrections from community members' would probably be too long and make the post slightly illegible.

erciccione:
> I know it makes it longer which is what we're trying to avoid, but I prefer not to be credited as sgp anymore where possible.

Yes, i was only referring to the community part, not to the change related to your nick.

@dEBRUYNE-1 

Perfect. Changing now.

dEBRUYNE-1 (approved):

SarangNoether (approved):

PR #1175, opened 2020-09-02
Suggest changes in en.yml (#1177 added to this PR)

lh1008:
Hello everyone, 

I suggest changing the word "is" for the word " has" in line-250 from the en.yml file. This to give a better contextualization to the text. Just read it with "is" and let me know your comments. 

Thank you.

:)


SarangNoether (approved):

erciccione (approved):
I'm fine with the change, but please consider that for any string change (even minimal like this one) contributors on Weblate, depending on their notifiction settings,  will receive a notification and the string will be marked as "need change" for all translations, which means translators will have to spend time to retranslate it.

Pointing it out because this is the third PR you make in a short time which fixes typos. Would be better to make one single PR, so that translators won't receive multiple emails/notifications for something so minimal. If you encounter typos during your translation work, please note them and then open a single PR after you are done. For the same reasons i explained above, also consider open PR that changes strings only when the change will make a strong difference :)

lh1008:
Thank you erciccione for your comment. I will be making a bigger Pull Requeste with the typos, and suggestions I find once I finish.

Thank you and will be adding Pull Request #1177 to this one.

erciccione (approved):
Thank you :)

PR #1178, opened 2020-09-05
Open links in the 'Merchants' page in new tab and add 'noreferrer'

erciccione:
This is useful to avoid "reverse tabnapping" attacks (the merchants page is  very exposed to those) and to limit profiling users  coming from getmonero who visit the merchants listed.

### DIFF
--- a/community/merchants/index.md
+++ b/community/merchants/index.md
@@ -5,7 +5,7 @@ permalink: /community/merchants/index.html
 ---
 {% t global.lang_tag %}
 <div class="merchants text-center container description">
-    <p>{% t merchants.intro1 %}<a href="https://github.com/monero-project/monero-site/issues"> {% t merchants.intro2 %}</a> {% t merchants.intro3 %}</p>
+    <p>{% t merchants.intro1 %}<a href="https://github.com/monero-project/monero-site/issues" target="_blank" rel="noreferrer noopener"> {% t merchants.intro2 %}</a> {% t merchants.intro3 %}</p>
 </div>
 <div class="merchants">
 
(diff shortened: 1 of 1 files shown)


l33d4n (approved):

community/merchants/index.md
```diff
@@ -5,7 +5,7 @@ permalink: /community/merchants/index.html
 ---
 {% t global.lang_tag %}
 <div class="merchants text-center container description">
-    <p>{% t merchants.intro1 %}<a href="https://github.com/monero-project/monero-site/issues"> {% t merchants.intro2 %}</a> {% t merchants.intro3 %}</p>
+    <p>{% t merchants.intro1 %}<a href="https://github.com/monero-project/monero-site/issues" target="_blank" rel="noreferrer"> {% t merchants.intro2 %}</a> {% t merchants.intro3 %}</p>
```
00-matt, line 8:
```suggestion
    <p>{% t merchants.intro1 %}<a href="https://github.com/monero-project/monero-site/issues" target="_blank" rel="noopener noreferrer"> {% t merchants.intro2 %}</a> {% t merchants.intro3 %}</p>
```

community/merchants/index.md
```diff
@@ -41,7 +41,7 @@ permalink: /community/merchants/index.html
             {% assign sortedMerchants = toplevel.merchants | sort: 'name' %}
             {% for merchants in sortedMerchants %}
             <div class="col-md-6 col-sm-6 col-xs-12">
-                <p><a href="{{merchants.url}}">{{merchants.name}}</a>
+                <p><a href="{{merchants.url}}" target="_blank" rel="noreferrer">{{merchants.name}}</a>
```
00-matt, line 44:
```suggestion
                <p><a href="{{merchants.url}}" target="_blank" rel="noopener noreferrer">{{merchants.name}}</a>
```

00-matt (commented):
Good catch. Although [the standard says that `noreferrer` implies `noopener`](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer), whenever I see it in use or in any recommendations, they seem to be together. Maybe some browsers don't follow the standard properly.

Also maybe there's a way to do this in Jekyll to all external links, instead of doing it manually where it could potentially be forgotten.

erciccione:
> Maybe some browsers don't follow the standard properly?

I assumed was the case some time ago and most of the documentation is simply outdated. I agree we can just add `noopener` and forget about it.

00-matt:
> AFAIK that's not possible with jekyll. We could write a ruby script
> to change all external links, but i think that would be overkill.

That's annoying :( -- although I don't think it's a big deal because
like you said in the first comment, the merchants page is the most
exposed. The risk would be much lower when linking to somewhere like
Reddit or a sponsor like MyMonero.

community/merchants/index.md
erciccione: Changed both. Thanks :)

PR #1185, opened 2020-09-12
View-only wallet guide: update instructions + create from private view key in GUI

rating89us:


erciccione (commented):
The snippet on the top of the guide for every other language will need to be changed to `{% include disclaimer.html translated="no" translationOutdated="yes" %}`. Could you do that? Otherwise i'll do it in another PR.

rating89us:
Updated

erciccione:
Sorry, i got confused. The snippet should be changed to `{% include disclaimer.html translated="yes" translationOutdated="yes" %}` if the guide was already translated in that language (so if `translated="yes"`). If the guide isn't translated, you can simply copy paste the new guide and replace the old one (keeping the snippet as it is: `{% include disclaimer.html translated="no" translationOutdated="no" %}`

rating89us:
Updated

erciccione (approved):
Looks good. Thanks :)

Issue #1188, opened 2020-09-14
Add video "Community & Key Aspects" to "What is Monero?" page


erciccione:
https://www.youtube.com/watch?v=lTdxUpkarjE

#1187 should be kept in consideration when adding this.

erciccione:
The 'Community' video is heavily outdated and there seems to be no plans to update it. Doesn't make sense to add it to the website. Closing this.

PR #1190, opened 2020-09-16
Reorganize Monero videos

### LINKED ISSUE #1187: Improve visibility of the videos in the "What is Monero?" page
Some [users report](https://www.reddit.com/r/Monero/comments/iaww0p/getmoneroorg_updated_improved_rss_feed_new/g57j2xq/) they are easy to miss. We could move them somewhere else or make them more visible.

erciccione:
Closes #1187

- Remove carousel structure in the 'What is Monero' page and leave only one video: 'Monero: the essentials'
- Move technology specific videos to their moneropedia entry: Ring Signatures, RingCT, Stealth Addresses
- Add aria-label to imrpove accessibility of the videos

If we like this change, i suggest to also create a FAQ entry saying something like "Are there some videos about how Monero works?" and then link to the videos. They are meant to be watched one after another and this new system kinda break that. I'm proposing it because i don't think people actually notice and watch those videos at the moment. Adding a faq with the list of videos in the right order will fix that and will allow us to link to all the videos at the same time.

We have one video missing here: #1188. We can decide where to put it once we have it.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -2,7 +2,7 @@ email: dev@getmonero.org
 name: Monero
 description: Monero is a digital currency that is secure, private, and untraceable.
 keywords: "monero, xmr, bitmonero, cryptocurrency"
-baseurl: ""
+baseurl: "https://www.getmonero.org"
 url: "https://www.getmonero.org"
 
 markdown: kramdown
--- a/css/custom.css
+++ b/css/custom.css
@@ -975,6 +975,19 @@ a {
     text-decoration: none;
 }
 
+/* Videos */
+
+.box-video {
+  padding: 0.5rem 0;
+}
+
+.box-video video {
+  width: 100%;
+  margin-left: auto;
+  margin-right: auto;
+  display: block;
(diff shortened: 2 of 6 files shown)


selsta:
A lot of unnecessary whitespace on my phone.

IMG_0652

erciccione:
@selsta what browser are you using on mobile? looks good to me using firefox and chromium's mobile previews. No problems on my phone as well.

selsta:
Safari.

erciccione:
Apparently Safari felt free to freak out when the video wasn't contained in a `div` with proper padding. Solved using the same structure used for the Moneropedia videos. I created a dedicated `videos` section in the `GENERAL` CSS section where to put the class used for these videos and I adapted the html accordingly. Everything seem to be working as expected now, on all tested browsers (Firefox, Chromium, Chrome and Safari. Thanks @selsta for helping with Safari).

Ready for reviews.

selsta (approved):

PR #1196, opened 2020-09-17
Add Nano S/X; capitalize Trezor Model T; remove wallet versions; add supported wallets

rating89us:
Changes:
- Add Nano S and Nano X models
- Capitalize Trezor **M**odel T
- Remove "supported since" wallet versions: it's useless information for most users (old versions don't work anyway because of hard forks)
- Add wallets that support each hardware wallet type
- Add/update links to support pages in Ledger and Trezor websites

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -186,9 +186,9 @@ downloads:
   showhash4: "Two guides are available to guide you through the verification process:"
   hardware1: The Monero community has funded a
   hardware2: Dedicated Hardware Wallet (Kastelo)
-  hardware3: which is now in progress. Moreover, since CLI 0.12.1 and GUI 0.12.3 Ledger has
-  hardware4: integrated Monero into their hardware wallets.
-  hardware5: Trezor model T supports Monero since version 0.14.1.
+  hardware3: which is now in progress.
+  hardware4: Ledger Nano S and Nano X are supported in Monero CLI, Monero GUI, and Monerujo wallets.
+  hardware5: Trezor Model T is supported in Monero CLI and Monero GUI wallets.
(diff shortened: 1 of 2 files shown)


erciccione (approved):

_i18n/en.yml
```diff
@@ -186,9 +186,9 @@ downloads:
   showhash4: "Two guides are available to guide you through the verification process:"
   hardware1: The Monero community has funded a
   hardware2: Dedicated Hardware Wallet (Kastelo)
-  hardware3: which is now in progress. Moreover, since CLI 0.12.1 and GUI 0.12.3 Ledger has
-  hardware4: integrated Monero into their hardware wallets.
-  hardware5: Trezor model T supports Monero since version 0.14.1.
+  hardware3: which is now in progress.
+  hardware4: Ledger Nano S and Nano X are supported in Monero CLI, Monero GUI, and Monerujo wallets.
+  hardware5: Trezor Model T is supported in Monero CLI and Monero GUI wallets.
```
erciccione, line 191: I'm thinking of a way to introduce these changes without breaking the present translations. This structure will look very weird with already translated pages. See [pt-br for example](https://deploy-preview-1196--barolo-time-757cf9.netlify.app/pt-br/downloads/index.html#hardware).

The easy solution would be to use new keys, for example `hardware_ledger` instead of `hardware4`, but that would mean that these strings will be inserted between translated strings, which is ugly. Another solution could be to reduce the amount of changes for each string, so that the structure can stay the same and there is no need to change the placeholders/keys.

rating89us: I would leave it as it is now, it is just a line break that is wrong. Translations will fix the line break, when they become available.

erciccione: Might take a long time before all translations are adapted, but on a second thought you are right, it's not really a big deal.

erciccione (approved):

PR #1200, opened 2020-09-18
udpate hashes, name, version for CLI v0.17.0.0

binaryFate:


selsta (approved):
Hashes match with the ones submitted to https://github.com/monero-project/gitian.sigs/pulls

PR #1202, opened 2020-09-19
Add banner for October 2020 network upgrade

erciccione:
I changed a bit the structure and added useful links. ~~Now the banner appears on all pages, not only the home.~~

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -16,8 +16,10 @@ global:
   outdated: The content of this page is outdated and the might be not valid anymore. Contact the community if you need support.
   outdatedMin: This page has been updated since the translation. You can use this version, but it may be incomplete. Please use the
   outdatedVersion: English version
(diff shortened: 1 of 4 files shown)


00-matt (changes requested):
Sorry to be so picky, but could we increase the contrast of the text in the banner? My eyes aren't so good, and I'm colourblind so it's almost unreadable for me unless I select/highlight the text. I think the CSS needs to change `.upgrade-content` and `.upgrade-content a`.

erciccione:
@00-matt That's not picky, thank you for letting me know. I increased the contrast until we reached the AAA rate in the WCAG scale. Let me know if it's ok now. I took the chance to also improve the CSS a bit.

00-matt:
Thanks :) Just waiting for Netlify to rebuild.

00-matt (approved):

erciccione:
The banner is only on the homepage now.

selsta (approved):

PR #1208, opened 2020-09-24
Refactor navigation menu to use i18n plugin

erciccione:
This continues the work of refactoring parts of the website that weren't adapted when introducing the i18n plugin, resulting in a lot of duplicate files, impossibility to scale and hard maintainability. This PRs fixes only the navigation menu, but we should also make the footers use the i18n plugin, so we can remove the `_data/lang` folder altogether. I will take care of that soon.

For clarity, the sections which still need to be refactored are:

- footer-1
- footer-2
- User guides
- Moneropedia entries

The last two are particularly difficult to refactor because a lot of user guides and Moneropedia entries where translated since the implemenation of the i18n plugin for the other half of the website. Adapting all the translated entries to the new system will be changelling and time consuming (see logs of the last community meeting for details).

Changes:

- Refactored navigation menu (in `_includes/header.html`) to make it use the i18n plugin. The navigation menu was previously not added in the i18n system and was unscalable and hard to maintain in that structure.
- Removed all duplicated navigation files of each language. This structure required to edit all of them (13 files) when a change was needed.
- This fixes missed entries and make possible to translate the navigation menu on Weblate
- Added the translations for all menu entries in all languages to their relative i18n/LANG.yml file
- Most of the menu entries in `navigation.yml` were duplicates of the 'titles' entries of the i18n/LANG.yml files. I copied the non-duplicated menu entries to a dedicate `navigation` section of the language fields. We use the entries in the `titles` section for the already existing menu entries.

PR #1217, opened 2020-09-30
Refactor footer to use i18n plugin 

erciccione:
Based on #1208 

This is the last part of the refactor (started with #1208), which allow us to remove the _data/lang folder with all its content and make the footer use the 18n plugin. Now only the user guides and the Moneropedia are outside the localization system.

- Removed the _data/lang folder and all its content
- Since most of the entries are duplicates of the 'titles' in LANG.yml, we use those. For the two missing ones we add a 'footer' section
- Edited the README


l33d4n (approved):

erciccione:
Rebased

PR #1221, opened 2020-10-03
Add missing slash (footer nav)

l33d4n:


erciccione (changes requested):
This is already fixed by #1217

l33d4n:
I even reviewed it :sweat_smile:

PR #1223, opened 2020-10-04
Translated 8 Moneropedia entries into German

Wobole:
The following entries were translated:
- Account
- Address
- Airgap
- Atomic Units
- Block
- Blockchain
- Bootstrap Node
- Bulletproofs


Wobole:
There's some links to other Moneropedia entries (starting with an @) which currently don't work in the preview. This will change as soon as the terminology within these articles is updated with the German terms.

PR #1228, opened 2020-10-07
Merchants: remove 'monerocardgames.com'

erciccione:
According to a report, deposits don't get credited: https://www.reddit.com/r/Monero/comments/j5s0jx/avoid_monerocardgamescom/

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -408,8 +408,6 @@
       url: https://minko.to
     - name: Monero Betting
       url: http://moneroeetsad6quaq4weauz5ndsx5jf6y3z5zwtckhudciapom26byid.onion/
-    - name: Monero Card Games - Provably fair Monero card games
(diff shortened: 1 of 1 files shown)


dEBRUYNE-1 (approved):

00-matt (approved):
In this case, I think a removal is fine (site operator is missing, TLS cert expired, ...).

For other cases though, where the site operator might dispute the scam claims, maybe we could put a warning triangle next to the link with a tooltip saying that some people don't like this merchant.

erciccione:
> maybe we could put a warning triangle next to the link with a tooltip saying that some people don't like this merchant.

That's tricky, because could give the feeling that we make some sort of vetting, when we actually don't. I think we don't have to list any merchant on this page, if a complaint about scammy behaviour is credible enough, i think it's fine to just remove the merchant. But that's just my opinion, another person could think differently.

PR #1248, opened 2020-10-16
Add FAQ item

ThomasBucaioni:
Add "how to get Monero for fiat" in the Faq


erciccione (changes requested):
This is already fixed by #1083

erciccione (commented):
I think that in this status this entry is mostly useless. Doesn't add anything new from the entry above. We should either incorporate this entry to "How can i get Monero" or give some fiat specific information about trading monero. Like for example the fact that it's legal in most countries, or something like that

ThomasBucaioni:
> for example the fact that it's legal in most countries

List of illegal/ban countries added

SamsungGalaxyPlayer:
I get that "how to get Monero for fiat" is a common question, but the difficult of answering it in a simple FAQ is too higher of a barrier for me to suggest its inclusion.

erciccione:
Yeah. Probably better to just expand the entry we already have.

ThomasBucaioni:
> Probably better to just expand the entry we already have.

The items are merged

_i18n/en.yml
```diff
@@ -330,9 +330,9 @@ faq:
   acontribute4: "The Outreach workgroup wrote a useful article to help newcomers:"
   q1: How does Monero have value?
   a1: Monero has value because people are willing to buy it. If no one is willing to buy Monero, then it will not have any value. Monero’s price increases if demand exceeds supply, and it decreases if supply exceeds demand.
-  q2: How can I get Monero?
-  a2: You can buy Monero from an exchange or from an individual. Alternatively, you can try mining Monero to get coins from the @block reward.
-  a2exchanges: "In past, you needed Bitcoin to buy Monero, but that's not the case anymore. You can directly trade Monero for national currencies (USD, EUR, GBP, etc) or other cryptocurrencies on many exchanges. Some require KYC (proof of identification); others do not, like decentralized exchanges. On this website is available a list of exchanges where it's possible to buy/sell Monero (XMR):"
+  q2xmrfiat: How can I get Monero?
```
erciccione, line 333: Do not change the key if the content hasn't changed. Otherwise it will force translators to retranslate the string.

_i18n/en.yml
```diff
@@ -330,9 +330,9 @@ faq:
...
+  a2xmrfiat: You can buy Monero from an exchange or from an individual. Institutional exchanges can have very higher fees than local exchanges, i.e. peer-to-peer transaction sites, but beware of scammers as usual... Alternatively, you can try mining Monero to get coins from the @block reward. Monero is legal is most countries, however, it is illegal or banking banned in the following countries: Algeria, Bangladesh, Bolivia, Cambodia, Canada, China, Colombia, Ecuador, Egypt, Indonesia, Iran, Jordan, Morocco, Nepal, Pakistan, Russia, Saudi Arabia, Taiwan, Vietnam.
```
erciccione, line 334: Where did you get this list of countries? and what do you mean with `it is illegal or banking banned`

Either way i don't think would be wise to add a list of countries would be wise. It would need to be constantly maintained or people could be misinformed.

erciccione (changes requested):

_i18n/en.yml
ThomasBucaioni: The list is removed, it was from [wikipedia](https://en.wikipedia.org/wiki/Legality_of_bitcoin_by_country_or_territory). Regarding Canada, the ban means: 

«Companies dealing in virtual currencies must register with the Financial Transactions and Reports Analysis Centre of Canada (Fintrac), implement compliance programs, keep the required records, report suspicious or terrorist-related transactions, and determine if any of their customers are "politically exposed persons."»

SamsungGalaxyPlayer:
You reference a page on the site; please link to it.

Replace answer with the following:

You can buy Monero from an exchange or from an individual. Exchanges are the most common way to buy Monero; there are compliant exchanges in most jurisdictions. Some wallets include functionality to easily buy Monero with fiat or crypto. Alternatively, you can mine Monero to get coins from the @block reward.

ThomasBucaioni:
It's corrected

erciccione (approved):

erciccione:
This will need to be "sqashed and merged"

PR #1277, opened 2020-10-30
Change telegram chat to one controlled by us

livinginformation:


fluffypony (approved):
Reviewed

fluffypony:
For context: there have always been 2 Telegram groups, but one seems to be getting out-of-hand, so this changes the recommended Telegram group to one that is more Monero-centric.

Issue #1090, opened 2020-07-21
Upload nginx config file


erciccione:
An old nginx.conf is stored in this repo '_config/nginx.conf', but it's outdated. We should upload an updated one, so that in case we lose access to the current server, a mirror could be set up quickly using the config file.

@binaryfate already stated he will upload the nginx.conf file to this repo. Opening this issue mostly for tracking purpose.

erciccione:
#1279

erciccione:
Resolved.

PR #1279, opened 2020-10-30
Add current nginx conf file in _config

binaryFate:


erciccione (approved):
Thanks!

This closes #1090

Issue #1270, opened 2020-10-24
Introduce i18n system for user guides


erciccione:
User guides and Moneropedia entries are currently translated manually one by one, which makes impossible to track the status of a translated file and to reliably update the entries. Furthermore, changing/adding an entry requires editing a huge amount of files. This was recently mitigated for the Moneropedia (#1198), but the system as a whole remains broken and unmaintanable.

About half website wasn't adapted correctly to the redesign, which was sloppyly implemented and, beside breaking important functionalities, was only partly adapted to the new i18n system: FAQ, footers, navigation menu, Roadmap, User guides, Moneropedia and Hangouts were left out for some reason or implemented uncorrectly. I've been fixing them all during this year and then i added them to Weblate. Now only user guides and Moneropedia are left. These two are much more complex to refactor, because during the years many translations were added. This hugely complicates things.

I've been thinking on ways to fix the system. The best option is to leave the markdown files as they are and use a converter to generate strings and serve them to weblate.

### Process
We will first introduce this for the user guides. After it will be well tested, we can adopt it for the Moneropedia as well.

The plan is to use `po4a`, a very versatile converter which we will use to extract the English text of a markdown document and put it into a tamplate file (`.pot`). Since we have documents already translated, we will have to extract the translated strings from these files, so that we don't need to retranslate them all. For this process to be successful, all files must have the same structure.

The extracted strings will be added to dedicated files in this repository and Weblate will pick them up. Translators will work on them on Weblate, which will commit the changes to these `.po` files. The translated strings will then be extracted from the .po files and transferred to the markdown documents, which will be picked up by Jekyll, as usual.

For a detailed overview of how po4a will work, see [this document](https://www.systutorials.com/docs/linux/man/7-po4a/).

Using this converter will allow us to have a system for translating markdown files, without making time consuming and invasive changes to the system we already use. In fact, jekyll will still work with the markdown files. The translation system will be separated entirely. Leaving the markdown files as they are will also make us ready for when Weblate will support markdown files (https://github.com/WeblateOrg/weblate/issues/3106) and using po4a won't be necessary anymore.

### What needs to be done

- [x] The open PR translating user guides (German: #1224 #1249 #1250 #1251 #1252 #1253 #1254 #1255. Norvegian: #1266 but more will come) must be merged/closed, so that it will be possible to work on the next point

- [x] All user guides in all languages must have their structure standardized and fit the English version.

[...]

PR #1282, opened 2020-11-01
User guides: New location and localization process for screenshots + clean all duplicated files

### LINKED ISSUE #1082: Fix insane number of duplicated png files used for user guides
While upgrading a user guide I realized that the files contained in /resources/user-guides/png are useless. All screenshots used by the guides are in _i18n/$LANG/resources/user-guides/png. The thing is that every language has its own png folder, but most of the screenshots are in english. This means we have a total of **1430**  (110 * 13 languages) pictures used for the user guides in the repository, when we could have only **110**. All screenshots (or at least the vast majority of them) are duplicates.

This is a huge waste of space. I believe that by fixing this issue we could save at least 120 MB. The current size of the entire repository is about 450 MB.

Jekyll doesn't handle I/O very well, but anything would be better thant the current system. I will investigate this issue further, but resolving it will require a deep check all user guides for each language one by one, so it could require some time.

erciccione:
This is a huge PR. We delete more than 1000 files; i strongly suggest to review it commit by commit.

~~Based on #1275, so please review that one first.~~

This PR changes the way we deal with user guides screenshots (closes #1082). Right now we have a `png` folder for every language, which contains the screenshots for every user guide for every language. The almost totality of these screenhsots are in English, so we have hundreds of files which are duplicates. We also have two folders where English screenshots are stored: `/_i18n/en/user-guides/png` and `/resources/user-guides/png`. These two folders were serving the same purpose, as a result some screenshots are updated in one and some in the other. A script in `_plugins` copys stuff around and make everything work.

I drastically changed and simplified the system. All screenshots are now in one single place: `img/resources/user-guides`. We only store one copy of all screenshots in the `en` folder and we use those for all languages. When somebody wants to add a localized screenshot for a user guide, they will only need to create a folder for the language and upload the screenhsots (instructions in the README). The results of implementing this new system will be:

- Less files moved around for no reason
- Avoid confusion generated by having the same screenshots in two different places. Now everything is cleaner and more immediate.
- It will be clearer which guides have screenshots translated (we only have the screenhsot for a language if it is translated)
- Removed a huge amount of duplicated files which had no purpose and only generated cluttering and confusion (>1000 images)
- The new i18n system for the user guides (#1270) will benefit from these changes.


selsta (approved):
looked over it as good as I can

PR #1304, opened 2020-11-12
FAQ: add Monero videos ("Are there videos I can watch to learn about Monero?")

erciccione:
Based on #1291

This PR is a follow up of #1190, as i suggest in the description:

> If we like this change, i suggest to also create a FAQ entry saying something like "Are there some videos about how Monero works?" and then link to the videos.

Pushing as draft because i want to check the preview on various browsers before opening it for reviews. Chrome and Chromium seem to attempt to download the videos when we click on the links, but apparently this is caused by local configurations. (Update: it is caused by a local configuration. The videos are shown correctly on chromium and Chrome on an Ubuntu VM. I'm opening this PR for reviews)


SamsungGalaxyPlayer:
image

The initial R in RingCT isn't capitalized for some reason.

I approve of everything else :)

erciccione:
@SamsungGalaxyPlayer That's unrelated to this PR and probably happens somewhere else. The script that create the Moneopedia links doesn't understand capitalization, so the link is always lower case.

SamsungGalaxyPlayer (approved):
Worth addressing that bug if possible since it does look unattractive, but I approve of this PR if it needs to be addressed elsewhere.

erciccione:
Opened #1352

PR #1312, opened 2020-11-14
Updating "Wallet/daemon isolation with Qubes + Whonix"

santorihelix:
**What I've changed:**
+ Added notice about increasing the daemon workstation's private storage and how to estimate the space needed.
+ Changed `monerod` location in the systemd file to `/usr/bin/monerod` from `/usr/local/bin/monerod` since that is where it is located now.
- Removed the parts about installing monero since it is now preinstalled in Whonix.
+ Changed mentions of `gedit` to `nano` because `gedit` is not installed by default.
+ Set outdated to False

I think the name should be changed to exclude the CLI part since the guide works for both monero-gui and monero-cli but to avoid mismatch between title and filenames I left it out for a later PR.

### DIFF
--- a/_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
+++ b/_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
@@ -13,7 +13,7 @@ This is safer than other approaches which route the wallets rpc over a Tor hidde
 
   - The first workstation will be used for your wallet, it will referred to as `monero-wallet-ws`. You will have `NetVM` set to `none`.
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://web.getmonero.org/downloads/#blockchain). Keep in mind that the blockchain will take more space with time.
+  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://web.getmonero.org/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
 
 ## 2. In the AppVM `monerod-ws`:
 
@@ -71,7 +71,7 @@ user@host:~$ sudo chmod +x /rw/config/rc.local
 
 ```
 user@host:~$ sudo mkdir /rw/usrlocal/etc/qubes-rpc
-user@host:~$ sudo gedit /rw/usrlocal/etc/qubes-rpc/user.monerod
+user@host:~$ sudo nano /rw/usrlocal/etc/qubes-rpc/user.monerod
 ```
 
 Add this line:
--- a/_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
+++ b/_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
@@ -13,32 +13,14 @@ This is safer than other approaches which route the wallets rpc over a Tor hidde
 
   - The first workstation will be used for your wallet, it will referred to as `monero-wallet-ws`. You will have `NetVM` set to `none`.
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`.
+  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://web.getmonero.org/downloads/#blockchain). Keep in mind that the blockchain will take more space with time.
 
 ## 2. In the AppVM `monerod-ws`:
 
-+ Download, verify, and install Monero software.
-
-```
-user@host:~$ curl -O "https://downloads.getmonero.org/cli/monero-linux-x64-v0.11.1.0.tar.bz2" -O "{{ site.baseurl_root }}/downloads/hashes.txt"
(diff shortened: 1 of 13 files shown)


_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
```diff
@@ -13,32 +13,14 @@ This is safer than other approaches which route the wallets rpc over a Tor hidde
 
   - The first workstation will be used for your wallet, it will referred to as `monero-wallet-ws`. You will have `NetVM` set to `none`.
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`.
+  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://web.getmonero.org/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
```
erciccione, line 16: Please change to the link to:
```suggestion
  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://{{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
```
This could be a useful trick to check the size of the blockchain, but keep in mind that that snapshot could be a bit back the current height. Still good to get an idea of the size though.

erciccione (commented):
Thank you :)

> I think the name should be changed to exclude the CLI part

As far as i know only the CLI is preinstalled on Whonix, so makes sense for now to keep this guide focused on the CLI.

These changes will need to be added to the other languages as well, but translators are at work on these files and we need to be careful to a couple of things. So, don't worry about it, i will take care of that later.

santorihelix:
@erciccione I committed the site.baseurl change.
> As far as i know only the CLI is preinstalled on Whonix, so makes sense for now to keep this guide focused on the CLI.
Actually as per [the Whonix wiki](https://www.whonix.org/wiki/Monero#Installation) monero-gui is also preinstalled. The list of monero related applications on Whonix can be seen in [this citation](https://www.whonix.org/wiki/Monero#cite_note-2). 
If you have some time I'm a bit curious on the translations part. Was it ever considered to keep the /resources/ separate for translations to allow quicker updates?

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
```diff
@@ -13,32 +13,14 @@ This is safer than other approaches which route the wallets rpc over a Tor hidde
 
   - The first workstation will be used for your wallet, it will referred to as `monero-wallet-ws`. You will have `NetVM` set to `none`.
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`.
+  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain](https://{{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
```
erciccione, line 16: Sorry, i don't know what i was thinking. This is the correct URL.

```suggestion
  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain]({{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
```

erciccione:
> Actually as per the Whonix wiki monero-gui is also preinstalled. The list of monero related applications on Whonix can be seen in this citation.

I didn't know! Better that way :)

> Yes I've noticed that. I didn't want to link third party sites because I'm not sure what the policy is on that so I added the "estimate" part.

I prefer the current way. You are right, it's better to not link to third party sites for these kind of things when we can avoid it.

> Are translators actively working on translating this at the moment?

There are two community funded translators currently working on the user guides. I would say let's make all the changes we want to make, i'll deal with making sure the translators are kept updated about the changes.

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
santorihelix: All is good now.

santorihelix:
> I would say let's make all the changes we want to make

So should I apply the changes excluding "the tip" to other languages? Honestly it would only take like 15 minutes and we could get all of them at least _working_ this way.

erciccione:
Fine for me to keep it as it is, but yeah the changes need to be added to the other languages if that guide is not translated yet. But not for German, so we avoid conflicts for the translators currently working on it. I'll take care of letting the translators know about the changes.

erciccione (commented):
The versioning snippet needs to be changed to `{% include disclaimer.html translated="yes" translationOutdated="yes" %}` for these languages: `ar`, `ru`, `nl`, `fr` and `es`.

We should probably merged this after #1270 is resolved.

santorihelix:
Hey @erciccione! I added commits changing those snippets fyi. Wouldn't it be easier for the translators to merge this before they finish translating so that they don't have to submit / edit their translations after this is merged?

erciccione:
> Wouldn't it be easier for the translators if this was merged before they finish translating so that they don't have to submit / edit their translations after this is merged? 

I'm trying to discourage changes to user guides and moneropedia at the moment. It's easy to lose track track of the changes and i want to make sure translators are updated and their work is synchronized. In this case the changes are mostly deletions and small changes, so we can go ahead and merge it.

>  Also, I can do the .po conversion and upload it if needed.

Thanks, but that's not necessary at the moment. We will proceed with the creation of the po files after all the translations are merged, to avoid problems of conflicting structures :)

erciccione (approved):

erciccione:
@wobole @MindWallet Please make sure you use this version of the guide when translating. If you already translated this guide, make sure to adapt it. DMing you.

PR #1313, opened 2020-11-15
Clarify ambiguous 'Ring Size' entry in moneropedia

binaryFate:
Current version is ambiguous, leading readers to think the ring size had no minimum up to version 0.13: https://www.reddit.com/r/Monero/comments/jtw48a/interesting_article_with_a_financial_slant_about/gc8hcwv


erciccione (approved):
This will need to be added to the other languages as well. I will take care of that after this is merged.

SamsungGalaxyPlayer (changes requested):
I don't like this new wording, since it implies Monero allowed only rings 11 and larger before 0.13. Revised below.

---

Ring size (or "ringsize") refers to the total number of signers in a @ring-signature. Before release [0.13.0 "Beryllium Bullet"]({{ site.baseurl }}/2018/10/11/monero-0.13.0-released.html), it was possible to select an arbitrary number of signers (sometimes with minimum limits). With release 0.13, it was decided to set this number to 11 to enforce transaction uniformity. 

# Ring Size History

| Release | Ring Size |
| --- | --- |
| Initial Release | All ring sizes allowed (0-decoy rings allowed) |
| 0.9.0 | Minimum ring size 3 |
| 0.11.0 | Minimum ring size 5 |
| 0.12.0 | Minimum ring size 7 |
| 0.13.0 | Mandatory ring size 11 |

erciccione:
ping @binaryFate

PR #1328, opened 2020-11-23
daemon-rpc guide: Fix target_height value for sync_info

erciccione:
See https://github.com/monero-project/monero/issues/7029#issuecomment-732232037


selsta:
17:14 <@moneromooo> Most people get it wrong. Target height is a sync mode variable. It has no meaning after you're synced.

selsta (approved):

PR #1343, opened 2020-11-28
Add additional unlock_time details to Moneropedia

SamsungGalaxyPlayer:
Requires a review by a knowledgeable individual, such as @TheCharlatan or @Mitchellpkt.


TheCharlatan (commented):
Thanks for adding this. I just have a suggestion for additional content.

_i18n/en/resources/moneropedia/unlocktime.md
```diff
@@ -1,13 +1,23 @@
 ---
```
ChanarinThintip, line 1: _i18n/en/resources/moneropedia/unlocktime.md #1343

Mitchellpkt:
Good updates, thanks for writing this up sgp.

Hmm, I'm not sure if this is in scope of the doc, but I'll mention it just in case - there is a mismatch between CLI & RPC behavior around transaction generation. If I recall correctly, the wallet only accepts height-like values (up to millions), but the RPC accepts any uint64 value (up to 18,446,744,073,709,551,615). Thus transactions with very high lock values were either submitted by RPC or custom software.

Mitchellpkt:
Oh, and also potential wording tweak, clarify
"_The integer should be an ***absolute*** block height value or Unix epoch timestamp, not a relative value_"
to
"_The integer value will be interpreted by the protocol as an ***absolute*** block height value or Unix epoch timestamp, not a relative value_"

SamsungGalaxyPlayer:
@Mitchellpkt @TheCharlatan please review when you have a chance

TheCharlatan (approved):

erciccione:
Let's wait a bit before merging this. #1270 must be kept in consideration.

plowsof:
closed by #2076

Issue #1290, opened 2020-11-06
Add Fork Networking to the list of merchants


apezio:
#### What's the name of your business?

Fork Networking, Inc.

#### What's the URL of the website?

https://www.forked.net/

#### Write a brief description of your business/service (max 50 characters)

Provides Dedicated Servers and Virtual Private Servers (VPS).

#### Enter the URL where a Monero mention can be found

https://www.forked.net

#### How do you accept Monero? *(choose only one option)*

- [X] Native wallet

#### How would you categorize your business? *(choose only one option)*

- [ ] Exchange
- [ ] Block Explorer
- [ ] Payment Gateway
- [X] Web Hosting
- [ ] Tool
- [ ] Service
- [ ] Goods
- [ ] Entertainment

#### Comment

Accepting Monero directly since 2017. Thanks!

erciccione:
When trying to access the website through TLS, i get a BAD_CERT error. Please fix. We already link to `https://www.forked.net/` in [Sponsorships](https://www.getmonero.org/community/sponsorships/)

erciccione:
ping @apezio

erciccione:
ping.

erciccione:
No answer. Closing.

apezio:
Sorry, fixed this awhile ago but didn't reply here.  I was missing an intermediate chain cert. sigh.

erciccione:
@apezio please edit the OP with the updated link. Note that it's still possible to access forked.net without TLS. That should be fixed.

apezio:
Thanks, I have fixed that now.

apezio:
#please edit the OP with the updated link.

@erciccione Done.

PR #1351, opened 2020-12-03
Sponsorships: Remove link to fork networking since their certificate is invalid

erciccione:
We can revert this commit once they fixed the problem. I already tried to contact them in #1290


selsta (approved):

apezio:
All fixed now.  SSL working as intended.

erciccione:
readded in #1589

PR #1355, opened 2020-12-05
Add 'synchronized' to 'get_info' call in daemon rpc guide

erciccione:
Added with https://github.com/monero-project/monero/pull/7030


selsta:
Note that this only got added to master. We should wait to include this until it is included in a release.

erciccione:
> We should wait to include this until it is included in a release.

https://github.com/monero-project/monero/pull/7161

selsta (approved):

PR #1366, opened 2020-12-12
Do not cache temporary redirects

codesoap:
Caching these locations led to the browser downloading older versions of Monero, when a link had been used before. 

For example: When I clicked the download link for the CLI-Wallet, Linux amd64 on https://www.getmonero.org/downloads/, Firefox downloaded `monero-linux-x64-v0.17.1.3.tar.bz2` instead of `monero-linux-x64-v0.17.1.6.tar.bz2`.

Note, that I never configured an nginx server before, so please double-check.

### DIFF
--- a/_config/nginx.conf
+++ b/_config/nginx.conf
@@ -46,59 +46,73 @@ server {
     # Downloads!
     location /win64 {
         rewrite ^ https://$server_name/cli/monero-win-x64-v0.17.1.1.zip redirect;
+        add_header Cache-Control "no-store, max-age=0";
     }
 
     location /win32 {
         rewrite ^ https://$server_name/cli/monero-win-x86-v0.17.1.1.zip redirect;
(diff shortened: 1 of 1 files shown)


erciccione:
Hi @codesoap thank you for the PR. Getmonero has been having caching issues for some time, but nothing that can be solved on this repository. The configuration is more complex than it appears and this patch alone is not enough to solve the issue. We should be able to resolve next week :)

codesoap:
> The configuration is more complex than it appears and this patch alone is not enough to solve the issue.

OK, thanks for the info.

> We should be able to resolve next week :)

Cool, looking forward to it 👍

erciccione:
Some changes to the nginx configuration have been added (not in this repository yet), but the problem still need to be solved.

> We should be able to resolve next week :)

I haven't been contacted by our back-end people yet. Hopefully they will soon.

I'm closing this PR in the meantime.

PR #1370, opened 2020-12-13
User guides: add a 'Back to the list' at the bottom of every guide

erciccione:


ChanarinThintip (commented):
Thanks

_layouts/user-guide.html
```diff
@@ -7,6 +7,9 @@
                 <div class="info-block text-adapt">
                     <div>
                         {{content}}
+                        <p style="text-align: right;">
```
rehrar, line 10: Generally not a fan of inline CSS styling, but I've been known to use it for little things. Other than this quibble, it's good to merge.

PR #1379, opened 2020-12-15
Added Feather wallet for Downloads Page

sh5x5j12:


erciccione (commented):
NACK

I think it's way too early to add Feather to the Downloads page. The project is still in beta and the first beta release was two months ago. It also needs extensive testing and vetting before we suggest it on Getmonero. Consider that the wallet we list have been around for > 1 year before getting listed.

sh5x5j12:
> 
> 
> NACK
> 
> I think it's way too early to add Feather to the Downloads page. The project is still in beta and the first beta release was two months ago. It also needs extensive testing and vetting before we suggest it on Getmonero. Consider that the wallet we list have been around for > 1 year before getting listed.

Thanks for reply, yes you're right it's in beta we can wait stable builds but isn't 1 year much?

erciccione:
> isn't 1 year much?

Consider that this software will handle people's money, i don't consider 1 year to be a lot. Even if we assume good faith because the developers are known, bugs and/or vulnerabilities can always happen. Before adding a wallet on that list we should be sure it has been tested and used long enough.

selsta:
IMO it make sense to continue this discussion once Feather reaches 1.0. They currently are in beta.

1 year sounds arbitrary to me.

erciccione:
> 1 year sounds arbitrary to me.

It is arbitrary :) My point is that it needs vetting and testing and that requires time.

erciccione:
@sh5x5j12 Could you close this PR? There is no point in leaving it open for now. We can reopen it once Feather will go out of beta or whatever people prefer.

PR #1382, opened 2020-12-20
Downloads: add link to release notes

erciccione:
Add 'release notes' link that points to the category "Releases" of the blog for both CLI and GUI releases.

Screenshot from 2020-12-20 14-35-32

Open to suggestions for a better place for this.


rehrar:
The placement is fine, I think.

PR #1383, opened 2020-12-22
Workgroups: add clarification about Monero Space being an LLC

### LINKED ISSUE #1380: Workgroup Page and For-Profit Workgroups
As far as I know, Monero Space is the first private, for-profit company wanting to be listed as a Monero Workgroup, so this is a new decision for getmonero to make. As such, it would make sense to me that a general policy be established in case other corporations would like to be listed in the future.

MoneroOutreach will need to do likewise, although we've tended to take our guidance from getmonero on matters of which things to list, wallets etc. This is a very tricky decision, we can imagine all sorts of future scenarios/actors in this permissionless space; I'm not exactly sure where the right balance is.

So that's in general.

More specifically this is a company that was started by Justin (and NeedsMoney?) after trying to turn Monero Community into an LLC and abruptly quitting when the community resisted. This is a continuation of that and putting it on the official Monero website is an endorsement of that.

erciccione:
Closes #1380

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -169,7 +169,7 @@ team:
   mrl_descr: As the name suggests, the Monero Research Lab is the workgroup focused on Monero related research. It's formed by scientists trying to find ways to improve Monero's protocol and features. Most of the protocol improvements that make Monero amazing, like CLSAG and @Bulletproofs, are the result of the work of this workgroup.
   mrl_start: Join the chatroom below, state your interests and your skills. Research is always ongoing and somebody will point you in the right direction.
   space: Monero Space
-  space_descr: Monero Space is an active, project-focused Monero workgroup that provides services to the Monero community. It hosts popular resources and organizes events such as Monero Meet.
+  space_descr: Monero Space is an active, project-focused Monero workgroup that provides services to the Monero community. It hosts popular resources and organizes events such as Monero Meet. It is incorporated in the US as an LLC (Limited Liability Company).
   space_start: The workgroup can be contacted by writing to their chatroom on Freenode.
   mrw: Malware Response Workgroup
   mrw_descr: The Monero Malware Response Workgroup provides resources and live support for multiple types of Monero-related malware.


rbrunner7:
A little comment: I became aware in the meantime that *LLC* already contains "company", being the abbreviation for *limited liability company*. So writing "LLC company", as I did, is doubling things.

Maybe not enough people know that *LLC* means, and it could be better to write:

> It is incorporated in the US as an LLC (limited liability company).

erciccione:
@rbrunner7 edited.

SamsungGalaxyPlayer:
***I significantly oppose this*** for all the reasons I mentioned in #1380. I allowed ErCiccione some discretion over their localization section (with wording I didn't completely feel was necessary), and Core was given some discretion over their section. I do not feel this description specifically under the Monero Space section aids anyone in any way except to fearmonger.

erciccione:
Your strong opposition of even mentioning the LLC is clear and makes this PR even more important.

MoneroArbo:
My 2c is I support this change

SamsungGalaxyPlayer:
I still don't think this is a good idea, but I will tolerate this if there is also an added sentence that the Community Workgroup is led by a US resident.

MoneroArbo:
I know a lot of discussion has been around the US jurisdiction, but for me it's the structuring as a for-profit company that is the issue. And I understand that the LLC hasn't really made any money. Probably the opposite. Nonetheless, I would be more comfortable if it were established as a non-profit entity.

erciccione:
@MoneroArbo @rbrunner7 or anybody from #1380: please approve this PR so we can queue it for merge. The vast majority of the people already voiced their approval, but would be good to have at least one "formal" approval.

MoneroArbo (approved):

lh1008:
I would just leave that it's an LLC if this is going through this way, pointing at something else is pretty bias, imo. 

I suggested a disclaimer because disclaimers are not only to warn but to delimit rights, limit responsibilities, legal liabilities, etc., disclaimers are open to definitions. 

That's why I pointed out to the user itself, a user could be a company, an individual, a country, a workgroup, etc. Disclaiming Monero's community's limited accountability (hope this is the right word; or responsibilities) when one user relates to another user would be better than pointing at something else. A user could be for e.g. relating to -> individual-to-individual, company-to-individual, company-to-company, country-to-country, country-to-individual, etc. (there are so many examples that it could be exhausting). I would address this information before merging, I know this is exhausting but think about it. 

Or like I suggested before, just build the tool, the money tool, Monero. Remove all companies, logos, etc., leave only the official wallets and documentation on how to use Monero. Whoever wants to use Monero will use the tool and that's it, but I know this might be "impossible" at this point, we all are users building a reputation, so at least addressing or limiting Monero's community's accountability between users is or might be the correct move here. Just think about it.

ProkhorZ (approved):

ProkhorZ:
@SamsungGalaxyPlayer A company has different rights and obligations than a natural person. Wasn't the main goal of Monero Space to make it easier to sign up users for third-party services?

You're doing a lot of special pleading on this issue. No problem, in the Christmas spirit we can forgive and forget, focusing on your record of positive contributions. But your use of the words _tolerate_ and _allow_ is out of order here. You don't have a veto. Waiting for complete consensus has never been the Monero approach.

We should go with the compromise supported by a majority, unless the Core Team wants to step in and make an exception for you.

selsta:
I personally don’t think adding this clarification is necessary here.

luigi1111:
See #1380

PR #1391, opened 2020-12-27
Add Monero Space meeting logs 2020-12-27

SamsungGalaxyPlayer:
Add Monero Space meeting logs 2020-12-27

### DIFF
--- a/_posts/2020-12-27-monero-space-meeting.md
+++ b/_posts/2020-12-27-monero-space-meeting.md
@@ -1,0 +1,450 @@
+---
+layout: post
+title: Logs for the Monero Space Meeting Held on 2020-12-27
+tags: [dev diaries, crypto]
+author: Justin Ehrenhofer (sgp_)
+---
+
+# Logs
+
+**\<sgp\_\>** Welcome to the very first Monero Space meeting!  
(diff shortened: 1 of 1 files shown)


erciccione (commented):
This raises the question: What logs should we add to getmonero?

Personally, i'm not a fan of adding logs as blog posts. Blog posts are very slow to load and every new one make the website heavier. Is this a one time addition or you plan to add logs weekly/biweekly?

I would be fine with a one time addition, but if becomes a habit for workgroups to add logs as blog posts, we will soon find ourself with a clogged and slow website.

erciccione:
Wait. Monero space has its own website. Why not post logs there?

SamsungGalaxyPlayer:
Kovri and the MCW have their own websites, and logs still got added here. I think if you read the logs, they are highly relevant to Monero. We can consider adding to our own site as well, that's probably a good idea.

I think if you want to remove logs from the getmonero website, you should open a new issue suggesting that logs (dev diaries) should no longer be a blog post type.

SamsungGalaxyPlayer:
Looks like there's some good reason to remove logs from the site entirely (I didn't know the build times were so long for text only), so closing.

PR #1260, opened 2020-10-19
Revamp of the 'Team' page (now renamed 'Workgroups')

erciccione:
Closes #895 

This changes completely the structure of the "Team" page, now renamed "Workgroups".

Instead of single members we show a list of the main workgroups, a description and brief instructions to start to contributing.

~~This PR is a draft because needs some minor edits and to give the opportunity to community member to improve descriptions and "where to start" sections~~. Feel free to suggest better wording, different instructions or new/different workgroups.

Preview:
Screenshot_2020-11-28  Workgroups

PR #1394, opened 2020-12-29
workgroups: change URL to /community/workgroups/

erciccione:
Now that #1260 is merged and live, we should change the URL of the Workgroups page from `/community/teams/` to `/community/workgroups/`.

@binaryFate we will need to set up a redirect.


SamsungGalaxyPlayer (approved):
LGTM assuming there's a redirect

binaryFate:
> Now that #1260 is merged and live, we should change the URL of the Workgroups page from `/community/teams/` to `/community/workgroups/`.
> 
> @binaryFate we will need to set up a redirect.


it's ready I'll enable it when we deploy.

PR #1401, opened 2021-01-02
Remove Shapeshift (.ai & .io)

Cactii1:
Shapeshift.ai no longer operating and Shapeshift.io de-listed Monero

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -83,10 +83,6 @@
       url: https://poloniex.com/exchange/btc_xmr
     - name: REXI.cc
       url: https://rexi.cc
-    - name: ShapeShift (Instant)
-      url: https://shapeshift.io/
-    - name: SideShift AI - A rapid coin swap that fully supports Monero
-      url: https://sideshift.ai
     - name: SimpleSwap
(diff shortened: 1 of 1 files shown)


_data/merchants.yml
```diff
@@ -83,10 +83,6 @@
       url: https://poloniex.com/exchange/btc_xmr
     - name: REXI.cc
       url: https://rexi.cc
-    - name: ShapeShift (Instant)
-      url: https://shapeshift.io/
-    - name: SideShift AI - A rapid coin swap that fully supports Monero
-      url: https://sideshift.ai
```
erciccione, line 89: Sideshift seems to be unrelated to shapeshift, and does list Monero.

erciccione (changes requested):

erciccione (commented):
Looks ok but the commit message is quite unclear. We are removing only shapeshift.io (not .ai). Something like `remove shapeshift.io and sideshift.ai`

Cactii1:
Ah jeese... sorry.. My bad, totally read that wrong. I need to be more careful.

PR #1406, opened 2021-01-05
MRL: remove '(English)' from page header

erciccione:
The page is being actively translated. The papers are not, but i guess people expect that, no need to specify anything.


rehrar:
Reviewed.

PR #1408, opened 2021-01-07
Merchants: add flipper

ashpumpkin:


erciccione (commented):
I don't know if we should include this in the Merchants page to be honest. The website is just a link to the telegram channel. People have no way to know if Monero is supported (we ask merchants to specify on their website that they support Monero) and no way to verify the bot actually works as advertised. Considering that Telegram is the favourite app for scammers, i'm inclined to NACK this Pr.

erciccione:
If there are no further comments, i'll just soon close this PR as rejected.

ashpumpkin:
@erciccione Hey, definitely not a scam, for now please reject. I will open a new one when we release a web version and mention monero on landing page

PR #1416, opened 2021-01-13
user guides: uniform structure of importing_blockchain.md for all languages

erciccione:
Part of #1270

- Remove author from the content of the guide
- Outdated translations were resetted (too time consuming to adapt them all and probably they will end up different anyway)


rehrar:
Reviewed

PR #1429, opened 2021-01-19
Removed merchants that aren't accepting monero

jarjarfan666:
These 3 merchants aren't accepting monero as a payment method

### DIFF
--- a/_data/merchants.yml
+++ b/_data/merchants.yml
@@ -121,8 +121,6 @@
 - category: Payment Gateways
   id: payment-gateways
   merchants:
-    - name: Bidali
-      url: https://www.bidali.com/
     - name: CDPay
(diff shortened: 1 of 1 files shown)


_data/merchants.yml
```diff
@@ -331,8 +327,6 @@
       url: https://drippinjunkies.co.uk/
     - name: Ecosystems retail - new & refurbished IT equipment
       url: https://store.ecosystems-group.co.uk
-    - name: FastTech
-      url: https://www.fasttech.com/
```
erciccione, line 335: Looks like they still accept Monero for payments (see the footer)

jarjarfan666: It's not an option in the checkout page:

fasttech

erciccione (approved):

PR #1432, opened 2021-01-22
roadmap: remove 'checked' from 2020

erciccione:
'checked' is used to specify which tab the user will be directed to when viewing the Roadmap. We use it only for the current year (2021)


jarjarfan666:
LGTM

rehrar (approved):

PR #1433, opened 2021-01-23
user guides: uniform structure of verification-allos-advanced.md, verification-windows-beginner.md, view_only.md, andbps_run_node.md

erciccione:
Part of #1270


wirrickelliot:
Should newlines be taken into account when comparing the structure of each file? I see some inconsistencies, but I'm not familiar enough with `po4a` to know whether they cause issues or not.

erciccione:
From my tests seems that the important is that paragraphs of texts are separated by a newline. Multiple newlines shouldn't effect po4a's parsing.

wirrickelliot (changes requested):
Remove the following lines (161-163) from _i18n/de/resources/user-guides/verification-windows-beginner.md:
```
Behalte die Auswahl von `Certify only for myself` bei und klicke auf `Certify`.

![kleo certify for self](png/verify_binary_windows_beginner/verify-win_kleopatra-certify-forself.png)
```

erciccione:
> Multiple newlines shouldn't effect po4a's parsing.

I had some doubts about this so i'm making more tests with different files and looks like the files have to match line by line (so mismatching newlines would be a problem). I'm looking into the best way to push the converted files. Let's put this PR in stand by, because might be superseded by another one.

erciccione:
Closing this. i will include these changes when creating the .po files of the related guides.

Issue #905, opened 2020-04-12
add "Transaction Weight" to Moneropedia


erciccione:
*Issue originally created by @stooge_coaster*

*This issue was created on gitlab and then migrated here. Only the original post was migrated, not the comments. Please take a look at the discussions on the original Gitlab issue before commenting here: https://repo.getmonero.org/monero-project/monero-site/-/issues/1033*

---

I'm suggesting ArticMine's description of transaction weight to be included in Moneropedia. His recent talk at 36c3 helps to elaborate on the transaction fee breakdown.

"Transaction weight is the metric used for an adjusting fee. Transaction weight is used to set the fee for a transaction" Refer to scalability to see how transaction weight is calculated.

Link to the scalability moneropedia post

[scalability](https://web.getmonero.org/resources/moneropedia/scalability.html)

PR #1461, opened 2021-02-09
Added transaction weight to Moneropedia.

tonyern:


tonyern:
I saw this on the issues page. It is issue #905. I made a basic layout of everything. I'm not that experienced with Monero and the blockchain to add anything to it so I'm hoping others can help add info as needed.

erciccione:
I think this is a good start, but i wouldn't create a Moneropedia entry for a couple of lines of explanation. Any idea on how to expand the entry?

tonyern:
Not sure how to expand it.

ProkhorZ:
Sources: the fee calculation is discussed in [Mastering Monero section 5.5.3.3,](https://github.com/monerobook/monerobook/blob/master/chapters/5.md) which may be outdated after changes in 2019. [ArticMine's talk can be found here](https://www.youtube.com/watch?v=btauHXDIM8M). I'll try to write a longer text if I understand this.

erciccione:
Note that thee fee structure is being revisited: https://github.com/monero-project/monero/pull/7819

ProkhorZ:
OK, the transaction weight isn't the most complicated part of the fee calculation or the dynamic blocksize. What do you think about the following text? I don't think we should include the precise calculation, which isn't done for other entries either. Relevant slide: https://i.imgur.com/MNw5Pda.png

"After @Bulletproofs significantly decreased the size of range proofs in Monero @transactions, the original size was still relevant for verification times. While proofs are relatively more efficient for a larger number of outputs, verifying each output will require a similar amount of time. To properly price the cost of verification, a 'clawback' is added to the transaction size of transactions with more than 2 outputs. The sum of the transaction size and this clawback is the transaction weight. For miners, the block weight is the sum of the weights of all transactions to be included in the block.

Not to be confused with Bitcoin's [weight units](https://en.bitcoin.it/wiki/Weight_units) introduced in the Segwit soft-fork, a different type of transaction weighting for different reasons."

ProkhorZ:
> Note that thee fee structure is being revisited: [monero-project/monero#7819](https://github.com/monero-project/monero/pull/7819)

Thanks for the warning. Looking at the discussion and changed files, I don't think this will affect the transaction weight or my proposed text. On the other hand, this suggests that explaining the transaction weight only explains a small detail of the fee calculation.

Issue #1460, opened 2021-02-09
Add <YogiWay> to the list of merchants


Peter-Liska:
#### What's the name of your business?

YogiWay

#### What's the URL of the website?

https://yogiway.cz/

#### Write a brief description of your business/service (max 50 characters)

yoga mats, bags and yoga accessories

#### Enter the URL where a Monero mention can be found

https://yogiway.cz/content/1-doprava-a-platba

#### How do you accept Monero? *(choose only one option)*

- [ ] Monero integrations *(github.com/monero-integrations)*
- [ ] Globee
- [X] Native wallet
- [ ] Other payment gateaway (Coinpayments, Morphtoken, etc. *Please specify*)

#### How would you categorize your business? *(choose only one option)*

- [ ] Exchange
- [ ] Block Explorer
- [ ] Payment Gateway
- [ ] Web Hosting
- [ ] Tool
- [ ] Service
- [X] Goods
- [ ] Entertainment

erciccione:
A correctly completed request, a rare sight! :P

Looks OK.

SydBal:
Cross posting a link to the PR for visibility: https://github.com/monero-project/monero-site/pull/1467

SydBal:
@Peter-Liska YogiWay has been added to the list of merchants. https://www.getmonero.org/community/merchants/

Peter-Liska:
Nice &#x1F44D;

PR #1467, opened 2021-02-12
[1460] Adding YogiWay data to merchants.yml

SydBal:
Referencing issue: https://github.com/monero-project/monero-site/issues/1460

Added the URL and site description to `monero-site\_data\merchants.yml`.


erciccione (approved):
Thanks :)

PR #1476, opened 2021-02-21
FAQ: Standardize video hyperlinks

### LINKED ISSUE #1188: Add video "Community & Key Aspects" to "What is Monero?" page

wirrickelliot:
Removed 'Monero:' from the Essentials link and added an i18n key for Ring CT.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -457,9 +457,10 @@ faq:
   avideos: >
     During the years the community has created a vast amount of informative content like articles and videos. Most of these videos are publicly available on platforms like YouTube. On this website we host a few videos that explain the fundamentals of Monero. To optimize their effectiveness, they should be viewed in sequence:
   video_intro: Introduction to Monero
-  video_essentials: "Monero: The Essentials"
+  video_essentials: The Essentials
   video_sa: Stealth Addresses
   video_ringsig: Ring Signatures
+  video_ringct: Ring Confidential Transactions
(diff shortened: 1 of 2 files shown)


erciccione (changes requested):
The video is from 2018, it's heavily obsolete, doesn't really make sense to add it in the current state. It mentions Kovri right at the beginning and again later in the video, saying that it masks IP addresses; the data shown is from 2018 and 2 of the 3 communication channels suggested in the video don't exist anymore (slack and MatterMost). At the very end, it goes into details in describing the FFS, which has been replaced years ago by the CCS.

Too bad, because this video has some interesting info for the viewer, but needs to be heavily updated.

wirrickelliot:
That's my mistake. I figured since the issue was opened only a few months ago that it was still something needing to be addressed.
If I remove the video, would it still be worth merging simply to unify the hyperlink text and add the i18n key for Ring CT?

erciccione:
I was under the impression that the "community" video was going to be updated. I asked about it on #monero-community and got no answer, so i'm assuming there are no plans to update it. In this situation doesn't make sense to add it.

> would it still be worth merging simply to unify the hyperlink text and add the i18n key for Ring CT?

Sure, but please squash the commits into one and change title and description of the PR.

erciccione (approved):

PR #1495, opened 2021-03-01
footer: fixed typo

salim-b:


erciccione (changes requested):
Thank you, but your commit description is too generic. Something like `footer: fixed typo` would be better.

salim-b:
@erciccione Ok, I've the changed the PR title accordingly.

erciccione:
You changed the title of the PR, but not of the commit, which still says only `Fix typo`. If you are using github's interface, i'm afraid the easier way to fix this is to close this PR, redo the commit and then open another PR.

salim-b:
@erciccione Happy now?

erciccione (approved):

PR #1496, opened 2021-03-01
Rings Cheatsheet version 20210301 added to Library section

baro77:
Following Addresses' one, I'm PRing a cheatsheet about Ring Signatures into Library section of the site.
I hope it's worth your attention,
baro


_i18n/en.yml
```diff
@@ -786,6 +786,9 @@ library:
   moneroaddressescheatsheet20201206: "Monero Addresses Cheatsheet (version 20201206)"
   moneroaddressescheatsheet20201206p: >
     Between the ‘big picture’ and the mathematical details, a single-page recap of various keys, addresses, scopes (private/public, spend/view, on-chain/off-chain, payer’s/payee’s) and their relations.
+  ringscheatsheet20210301: "Rings Cheatsheet (version 20210301)"
```
erciccione, line 789: I would not put the version. Just for readability, not a hard request

erciccione (commented):
I'm asking folks from the MRL to take a look :)

_i18n/en.yml
baro77: yep I agree, at least until next version will appear (if ever ;) )... so I would delete version from Addresses Cheatsheet as well...
By the way, in your opinion, what about making a "Cheatsheets" area between "Books" and "Magazines"? If you like it and it was my duty as proponent, can you provide some hints? (I'm worried to make a mess acting on site structure)

erciccione: > until next version will appear

I'm not sure would be necessary even in that case. You put the version number inside the PDFs, i would say that's enough.

> I would delete version from Addresses Cheatsheet as well

yes, i'm taking care of that right now.

> what about making a "Cheatsheets" area between "Books" and "Magazines"?

Yes that's a good idea. I'll take care of that too after this PR is merged.

UkoeHB (approved):

erciccione:
@baro77 your second commit is partially redundant with your first commit. I think would be just easier to remove it. It's not a problem to have a second PR :)

baro77:
>  I think would be just easier to remove it.

@erciccione uhmm ok I could "git reset" .... but in that case I would end up again with version in Rings Cheatsheet (I have seen #1497 removes the version only from Address Cheatsheet...).. is that ok? In that case are you going to take care of it as well?

erciccione:
IMO the easiest way to proceed is:

1. Remove your last commit (`git reset --hard HEAD^`)
2. Remove version number from "Rings Cheatsheet". After doing that, add this change to your first commit (`git commit -a --amend --no-edit`)
3. Now force push the new commit: `git push -f origin RingsCheatsheet`

This will remove the redundant commit and the version number from the Rings Cheatsheet, but without creating a new commit. #1497 will remove the versioning number from "Monero Addresses Cheatsheet", so that both will have no version number.

baro77:
> IMO the easiest way to proceed is:

Done! Thanks, I'm never comfortable rewriting history, and I didn't know "--amend" option of commit.. cool (y)

erciccione (approved):
Thank you. These graphics are very cool and i think are of great help :)

PR #1497, opened 2021-03-02
library: remove version from the 'Monero Addresses Cheatsheet'

erciccione:
https://github.com/monero-project/monero-site/pull/1496#discussion_r585397642


baro77:
not knowing about merging orders, to avoid possible need for manual merging I have removed Addresses Cheatsheet version it in the new commit for #1496 as well (the commit in which I have removed the version for Rings Cheatsheet)

wirrickelliot (approved):

PR #1448, opened 2021-02-02
daemon-rpc: update output of all calls and their description

erciccione:
I tested all the current calls and updated their description in our RPC guide for the daemon.

~~This is still a work in progress, i tested all the calls but many definitions need to be updated. Any help would be highly appreciated.~~

~~grep `TODO` to find the definitions that still need to be updated.~~

part of #1434

Issue #1434, opened 2021-01-24
Many RPC commands have been renamed or delated


erciccione:
We need to make a thorough refresh of the dev guides to check which commands have been renamed/deleted/added.

https://www.getmonero.org/resources/developer-guides/daemon-rpc.html
https://www.getmonero.org/resources/developer-guides/wallet-rpc.html

erciccione:
daemon-rpc is PRd in #1448. wallet-rpc is in progress.

ndorf:
> wallet-rpc is in progress.

Is someone actively working on this? I'd like to help, would be nice to avoid duplicating effort.

erciccione:
@ndorf not yet. I plan to start after #1448 will get some attention, but if you want to take care of it already. Sure, go ahead :)

erciccione:
Actually, please let me know if you intend to start working on the wallet-rpc user guide. So i can assign this issue to you as well.

ndorf:
I'm happy to help with the wallet, daemon or both guides. Please assign whatever you feel like to me.

erciccione:
Awesome, then would be great to start by providing the definitions for #1448, if you can :)

erciccione:
Note that more calls have been added since #1448. So after that's merged we should take another look and see what's missing.

erciccione:
We can consider this resolved.

PR #1499, opened 2021-03-03
wallet-rpc: Remove nonexistent parameter

ndorf:
The verbose flag on the incoming_transfers request doesn't actually exist, and
the key image is always returned. See https://github.com/monero-project/monero/blob/b8f3e44a3f7af299c3d6fbf2d21b78d8c99d6d7d/src/wallet/wallet_rpc_server.cpp#L1851


erciccione (approved):
Thanks. Note that all the calls on this page need to be refreshed (#1434), like already done for daemon-rpc (#1448).

Issue #1484, opened 2021-02-27
Revise "what is monero" page


SamsungGalaxyPlayer:
The current "What is Monero" should go further in explaining the most important qualities that set Monero apart from other assets.

***Note: the proposed revision should be mostly judged for content, not for style. It needs a further redesign. I'm not good at that part.***

Notice that the proposed version very clearly focuses on Monero's advantages in a way that immediately convey that Monero is not a shitcoin. It also contains calls for actions at the bottom:

"Join Community": https://www.getmonero.org/community/hangouts/
"Get Monero": https://www.getmonero.org/community/merchants/#exchanges
"Show me the full details": https://www.getmonero.org/resources/about/

### Current

![image](https://user-images.githubusercontent.com/12520755/109370444-eee5ce00-7865-11eb-987f-fc15ed854816.png)

### Proposed

![image](https://user-images.githubusercontent.com/12520755/109370432-e097b200-7865-11eb-8a15-1c46b9bfc4e9.png)

We can leave the video card at the bottom. I'm indifferent about that.

Conflicts with #1480. Related to #1483 #1485 #1486

robby-dermody:
as a possible alternative, this page could also be linked from the homepage only (e.g. the "Learn why it's the standard"), and keep the current What is Monero page (possibly with certain revisions that can be addressed separately)?

erciccione:
The suggestions you are proposing are basically entire redesigns and some would require long discussions before being implemented. I agree with some of the changes, but implementing them would take a lot of time and resources

Beside, some of these pages were refreshed not long ago and the text changed. I wouldn't make another (more intrusive) redesign any time soon.

erciccione:
On the merit of this issue, i like the idea of adding engagement buttons, but beside the fact that the text of the page is being changed by #1481 (so shouldn't be changed again any time soon), i don't like the content of those bullet points. They don't really explain what Monero is (which is the point of the page) and some of the content seems to be redundant with the 'About' page (points 3 and 4).

tallest-man:
@erciccione would you be able to provide some additional comments on your reasoning for:
"Beside, some of these pages were refreshed not long ago and the text changed. I wouldn't make another (more intrusive) redesign any time soon." and "(so shouldn't be changed again any time soon)"?

[...]

PR #1506, opened 2021-03-10
what is monero: add action buttons and remove redundant header

erciccione:
Requires #1481

Adding some action buttons as suggested in #1484 and removing redundant header.


wirrickelliot:
Do you think it would work to swap the 'About' and 'Get A Wallet' buttons? The first two paragraphs go together, so it might encourage users to read both of them before redirecting to learn more.

erciccione:
> Do you think it would work to swap the 'About' and 'Get A Wallet' buttons?

I have no strong opinion about this. I'm ok with swapping them.

wirrickelliot (approved):

PR #1508, opened 2021-03-13
Improve performance of some pages by lazy loading images

erciccione:
Lazy loading makes web pages load images only when they are actually
needed (when they need to be displayed to the user), thus saving
resources for both host and users and improving performances for the
latter.

| Page | Page weight before (load time) |  Page weight after (load time) | Space savings (time savings)
|---|---|---|---|
|index.html | 474 kB (546 ms) | 214 kB (373 ms) | -54% (-31%)
| get-started/what-is-monero/index.md | 320 kB (548 ms) | 246 kB (525 ms) | -23% (-4%)
| get-started/accepting/index.md | 254 kB (747 ms) | 193 kB (536 ms) | -24% (-28%)
| downloads/index.md | 397 kB (595 ms) | 243 kB (383 ms) | -38% (-35%)

(tests ran locally)

This should work our of the box on Firefox and Chrome.

https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -48,7 +48,7 @@ permalink: /downloads/index.html
                 <i>{% t downloads.gui_intro %}</i>
             </div>
             <div class="col-md-7 col-sm-12 align-blocks">
-                <img class="screen" alt="{% t accessibility.guiscreen %}" src={% if site.lang == "en" %}"/img/downloads/gui.png"{% else %}"/img/downloads/{{site.lang}}/gui.png"{% endif %}>
+                <img class="screen" loading="lazy" alt="{% t accessibility.guiscreen %}" src={% if site.lang == "en" %}"/img/downloads/gui.png"{% else %}"/img/downloads/{{site.lang}}/gui.png"{% endif %}>
             </div>
             <div class="col-md-5 col-sm-12 align-blocks">
               <ul class="logo">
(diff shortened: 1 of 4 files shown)


downloads/index.md
```diff
@@ -299,10 +299,10 @@ permalink: /downloads/index.html
                         </div>
                         <div class="row mob-wallets center-xs">
                             <a class="ext-noicon" href="https://support.ledger.com/hc/en-us/articles/360006352934-Monero-XMR-"><img src="/img/ledger.png" alt="ledger logo"></a>
```
wirrickelliot, line 301:
```suggestion
                            <a class="ext-noicon" href="https://support.ledger.com/hc/en-us/articles/360006352934-Monero-XMR-"><img src="/img/ledger.png" loading="lazy" alt="ledger logo"></a>
```

wirrickelliot:
I think eager loading might actually be preferred here to avoid any CLS issues. You might not agree with me, but I definitely think it's something worth considering.

wirrickelliot:
I believe best practice is to eager load images above the fold.

downloads/index.md
erciccione: Thanks. Added :)

erciccione:
I'm not sure it would be beneficial. The point of lazy loading is to show immediately the important stuff and the rest as soon as possible. We have few images in these pages, if we eager load the first 1 or 2, i'm afraid the performance improvements wouldn't be as significative.

wirrickelliot:
I only mentioned it because it's recommended on [this page](https://web.dev/lazy-loading-best-practices/#wrong-fold), but since we don't use any JavaScript, it might not be an issue.

wirrickelliot (approved):

wirrickelliot (approved):

PR #1509, opened 2021-03-14
reduce CLS when lazy-loading images

wirrickelliot:
Requires #1508 

Adds width and height attributes to lazy-loaded images. Allows browsers to calculate the aspect ratio and allocate adequate space to avoid layout shifting.

https://www.youtube.com/watch?v=4-d_SoCHeWE


erciccione:
Please rebase on #1508 (had a small conflict after the recent merges)

PR #1518, opened 2021-03-19
merchants: further clarify that we don't do any vetting of the merchants

erciccione:
slightly edited wording and highlight some text in bold

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -122,7 +122,7 @@ hangouts:
     monero-site: Where the development of this website is coordinated
 
 merchants:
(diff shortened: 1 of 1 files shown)


wirrickelliot:
It might be better to use a banner (similar to pages that need translation) since a good number of users might not even bother reading the text at the top. It would also help with situations where we to link to specific categories, and the top of the page isn't visible.

image

wirrickelliot:
Perhaps it would also help to put national flags next to each merchant to indicate what country they're based out of.

wirrickelliot (approved):

erciccione:
A banner could be a good idea, but let's merge this for now.

>

PR #1534, opened 2021-03-28
remove style from scroll-to-top button when focused via the mouse

wirrickelliot:


erciccione (approved):
Side note, but for some reason i still get the button coloured orange if i change tab and then come back to the page. This seems to happen only on Firefox though. No reason to stress about it. LGTM

PR #1542, opened 2021-04-02
Roadmap Kastelo Update

sh5x5j12:
I moved kastelo to the future section and changed into "upcoming".

It seems there's no development for it: 
https://github.com/monero-project/kastelo/commits/master


erciccione:
Honestly, maybe makes sense to just remove it. Development seems to have stopped long ago and there doesn't seem to be plans. No reason to have it in the roadmap at this point.

erciccione (changes requested):
I would remove it only from the roadmap. At least until we hear soimething from the devs.

erciccione (approved):
Needs to be squashed

PR #1544, opened 2021-04-04
home: align moneropedia card

alreadyburnt:
Fix card alignment.

### DIFF
--- a/index.html
+++ b/index.html
@@ -113,7 +113,7 @@ permalink: index.html
             <!-- END left two-thirds block-->
             
             <!-- right one-third block-->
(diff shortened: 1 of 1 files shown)


wirrickelliot (commented):
Looks good although I would suggest starting your commit message with "home:" since there are multiple index.html files. And then change the PR title to match.

wirrickelliot:
Please update the commit message as well.

alreadyburnt:
These are my first days with git so I don't really know how it works. There seem to be no such option from the web interface so I'm trying now from the cli. I hope not to mess up haha.

wirrickelliot (approved):
Thanks :)

erciccione (approved):

PR #1546, opened 2021-04-05
faq: add period to description

alreadyburnt:


erciccione (approved):
I guess it's fine for consistency, since we put a dot in the other descriptions.

luigi1111:
Disagree. If you (=anyone) want to make it a sentence, go ahead. The other intros I looked at were at least sentences, and even if they weren't, I'd rather remove the errant period than add it here.

alreadyburnt:
I'm not sure such a change is required just to add the period. However if u think so here's my proposal.

luigi1111 (approved):
Thanks.

PR #1547, opened 2021-04-05
hangouts: fix mailing list card alignment

alreadyburnt:


erciccione (approved):
This is not really necessary, i don't know if makes sense to add it.

alreadyburnt:
@erciccione I do not agree. All pages have the same amount of vertical whitespace between cards except this one, which makes it ugly and annoying.

Screenshot_2021-04-04  The Monero Project (1)
Screenshot_2021-04-05  Monero Research Lab (1)
Screenshot_2021-04-05  Sponsorships
Screenshot_2021-04-05  User Guides

PR #1551, opened 2021-04-05
meta: wrong icon path

alreadyburnt:
I'm not sure what I'm touching here, but these paths seem to be wrong.

screen


erciccione (approved):
from wirrickelliot: `I'm pretty sure they're used when adding a website to the home screen on android

PR #1555, opened 2021-04-06
hangouts: fix typo

alreadyburnt:


erciccione (changes requested):
This is incorrect. Moneroj is the plural of Monero. With `Moneron` the author means "person who uses Monero".

alreadyburnt:
Um I checked the translations and it seems you're right. The N key is next to J on the keyboard, so I assumed it was a typo.

PR #1558, opened 2021-04-06
contributing: reworded for contacting/joining workgroups

secure-designer:


wirrickelliot:
While I can agree that the original wording is a little unconventional, I think it's meant to emphasize that people can find ways to contribute that best suits their skills.

secure-designer:
I think affine might be too uncommonly used in that way so as to be clear.

https://www.dictionary.com/browse/affine doesn't show it fits.
https://www.merriam-webster.com/dictionary/affine does show further down on the page it's taken to mean "having affinity" but people might have a hard time finding that.

wirrickelliot:
Maybe something as simple as this this could work then: `"Join the workgroup that best suits your skills."`

secure-designer:
I'd be keen for `workgroup(s)` but then it would be `"Join the workgroup(s) that best suits (suit) your skills."` which is a bit awkward.

Maybe `"Join the workgroup(s) that your skills suit best."`?

erciccione:
I think we can keep it simple and singular. It's assumed that if you want to join more than one workgroup, you can. I'm fine with `Join the workgroup that best suits your skills.`

SamsungGalaxyPlayer (approved):
I like the wording "Contact or join the workgroup(s) to which you'd like to contribute."

erciccione:
> I like the wording "Contact or join the workgroup(s) to which you'd like to contribute."

Fine for me as well. LGTM

PR #1591, opened 2021-04-26
remove bitqist from the list of exchanges

rxdunlop1:
According to Bitqist via [a tweet from 21 May 2020](https://twitter.com/bitqist/status/1263438451715854336), they have been absorbed by another exchange called Bitvavo and no longer offer services.


erciccione (approved):
Thanks :)

PR #1592, opened 2021-04-26
make onion link point to the currently visited page

### LINKED ISSUE #1571: add the feature to conserve the page when converting to .onion with ".onion available" button on the Tor browser.
I couldn't come up with a better title.

image

erciccione:
onion.html now points to the page the user is on. The onion-location
button will point to this page instead of the homepage.

closes #1571


selsta (approved):
Looks ok, but untested.

PR #1609, opened 2021-05-06
MRL: add note to arcturus paper (was retracted)

erciccione:
The preprint has been withdrawn (https://eprint.iacr.org/2020/312), commenting it out until it's reuploaded or a new source is provided. Or maybe we should remove it completely? @SarangNoether?


SamsungGalaxyPlayer:
Would it be better to start with `RETRACTED:`?

erciccione:
don't know if it worth it, since users would be able to only read the introduction, without having access to the paper.

selsta:
> without having access to the paper.

It is still possible to read the paper by clicking on "All versions of this report"

erciccione:
> It is still possible to read the paper by clicking on "All versions of this report"

Then definitely makes more sense to just add a note.

selsta (approved):

PR #1617, opened 2021-05-10
update gemfiles, CI and netlifhy settings

erciccione:
Jekyll and its dependencies are heavily outdated. Wasn't easy to  find a working configuration, but this seems to do the job. Note that there are a lot of differences between jekyll 3 and 4, including much faster builds and improved caching (that's why i added `.jekyll-cache/` to the `.gitignore` ). I also needed to add the `webrick` gem, that now is requested as dependency, and to remove `rubysl-rexml`, which is not needed anymore.

~~For some reason, i'm unable to build the website using bundler after updating. Seems to build only when using `jekyll serve` directly. Pushing as a draft because this needs to be tested more before being merged and i wanted to use the buildbot to see if i can reproduce the problem with bundler.~~

CI and Netlify updated as well (commits were in #1620)


selsta (approved):
Netflify fails. Is this ok? Otherwise looks good.

erciccione:
yes, this PR updates the minimum version of ruby for netlify to 2.7. 2.6 (the current) won't work with new jekyll versions.

PR #1620, opened 2021-05-12
update ruby version in CI and netlify's settings

erciccione:
Requires #1617

The updates from #1617 won't work on old ruby versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


selsta:
3.0.x seems to fail. Also you might want to change actions/setup-ruby to https://github.com/ruby/setup-ruby

erciccione:
> 3.0.x seems to fail. Also you might want to change actions/setup-ruby to https://github.com/ruby/setup-ruby

Changed. Thanks.

I'm closing this and add the 2 commits (squashed into 1) to #1617, to make testing easier.

PR #1634, opened 2021-05-20
Add email contact to the top of the MRL page

SamsungGalaxyPlayer:


erciccione (changes requested):
Could you put the string in its own placeholder?

PR #1642, opened 2021-05-23
Home: Performance and SEO improvements

diegoramosxyz:
This commit fixes the following issues:

- Footer links do no have enough space between them: [Tap targets are not sized appropriately](https://web.dev/tap-targets/?utm_source=lighthouse&utm_medium=devtools)

- Certain images do not have dimensions: [Optimize Cumulative Layout Shift](https://web.dev/optimize-cls/?utm_source=lighthouse&utm_medium=devtools#images-without-dimensions)

- [Ensure text remains visible during webfont load](https://web.dev/font-display/?utm_source=lighthouse&utm_medium=devtools)


css/custom.css
```diff
@@ -2864,7 +2866,7 @@ footer a.white:hover, a.white:focus {
 }
 
 .footer-wrapper ul {
-    line-height: 1.7;
+    line-height: 3;
```
erciccione, line 2869: This looks quite ugly. Maybe better change it only for mobile?

also:
tordis

erciccione: ping @returndiego

returndiego: I fixed the issues you mentioned, @erciccione. I was unable to run the web server locally, so I made these changes sort of blindly. 

I will consider contributing more if the project is moved to a JavaScript/TypeScript platform such as [11ty](https://www.11ty.dev/). Ruby creates so much friction! Is there a good enough reason to use Ruby over Node.js when so much on the web uses JavaScript?

erciccione: > I was unable to run the web server locally

Did you take a loook at the instructions in the README?

> Is there a good enough reason to use Ruby over Node.js

The point is not the language used but the website generator. We used jekyll from the beginning, which is one of the most popular website generators

> when so much on the web uses JavaScript?

That's not necessarely a good thing. Getmonero doesn't use any javascript.

erciccione (changes requested):
Please squash your commits into only 1

erciccione:
ping @diegoramosxyz

PR #1652, opened 2021-05-30
replace freenode with libera in workgroups and hangouts

erciccione:
see https://github.com/monero-project/meta/issues/575

Opening as draft because it's not clear to me which channels are already bridged and which are not (https://github.com/monero-project/meta/issues/575#issuecomment-850951126)

PR #1662, opened 2021-06-08
remove references to mattermost

erciccione:
requires #1652

MatterMost is deprecated since we started using matrix.

- removed button from hangouts and replaced with matrix
- removed links from workgroups


selsta (approved):

steverusso:
Should we also update / remove:
* [the 'Mattermost' link in the footer](https://github.com/monero-project/monero-site/blob/8352829962bbc673d932917de8ac4a2c10a813c1/_includes/footer.html#L22)?
*  the 'MatterMost' references in the README?

erciccione:
> Should we also update / remove:

Done. Thanks.

PR #1670, opened 2021-06-11
create tools page

### LINKED ISSUE #1638: Remove the 'Tools' section from the Merchants page
The Merchants page lists businesses and services that accept Monero in exchange for said services/products. The 'Tools' section is an exception, as it lists tools like block explorers, which accept Monero donations.

I think the merchants page is not the best place for these services. I suggest to remove the 'Tools' section from the merchants page and instead create a dedicated section somewhere else.

steverusso:
Closes #1638.


erciccione (changes requested):
Some comments:

Don't use the structure of the merchants page. It's very ugly and it's not possible to translate it. Better come up with something new.

No need to create a dedicated `tools.yml` file, which cannot be translated. Better use `en.yml` and change the structure.

Have you checked that the tools listed are still existent/active?

steverusso:
> Better use en.yml and change the structure.

I moved the tools data into `en.yml`, but what do you mean change the structure?

> Have you checked that the tools listed are still existent/active?

I intended to in a follow-up PR, but I can just do that in this one if you want.

> ... the merchants page. It's very ugly and it's not possible to translate it.

I agree with this, but at the same time I think it should be consistent with similar pages on the site, and I defaulted to the merchants page because it does a good job of: "list links in groups." Thoughts?

erciccione:
> but what do you mean change the structure?

The structure you copied from the merchants page is not compatible with the localization plugin. See the structure of rest of en.yml to see what i mean.

> I intended to in a follow-up PR, but I can just do that in this one if you want.

Makes no sense to add a new section with outdated info. So yes, all info in this PR must be already up to date. Don't worry too much about that for now, this PR needs a complete rework.

> I agree with this, but at the same time I think it should be consistent with similar pages on the site, and I defaulted to the merchants page because it does a good job of: "list links in groups." Thoughts?

We have many structures we could use that would still be consistent with the rest of the website. The "Merchants" page is without doubt the worst page we have. It's ugly and not very functional. Let's not reuse that structure :)

A good start would be to categorize all the tools we want to add here and then decide how to display them. I think having only 2 sections: "Block Explorers" and "Miscellaneus" is not very useful. I also noticed you haven't added anything from the monero_tools guide.

steverusso:
> Better use en.yml and change the structure.

I think this is correct now :crossed_fingers: 

> We have many structures we could use that would still be consistent with the rest of the website.

I opted for the structure that is used for the work groups, user guides, and developer guides pages (you were right -- I certainly like this much better).

> I think having only 2 sections [...] is not very useful.

Agreed. I broke them into five sections: `Block Explorers`, `Address Generators`, `Network`, `Market` and `Miscellaneous`.

> I also noticed you haven't added anything from the monero_tools guide.

I brought over all of the tools from the `monero_tools` guide as well as the translations that were available for them. I then removed the `monero_tools` guide files.

> So yes, all info in this PR must be already up to date.

I checked all of the links on this new Tools page and found the following (just mentioning here before getting the ok to do anything regarding them):

1. [moneronodes.i2p.xyz](http://moneronodes.i2p.xyz/) can't be reached.
2. [nestorgames.com](https://www.nestorgames.com/) I think better belongs somewhere on the Merchants page?
3. [monerobase.com](https://monerobase.com/) seems to not have been updated since Dec 2018.
4. [xmrnotify.com](https://xmrnotify.com/) redirects to `https://darknetpages.com/monero-guide/`

---

Should I remove the `monero_tools` weblate `.po` files? (`_i18n/[lang]/resources/user-guides/weblate/monero_tools.po`)

erciccione:
This structure is definitely better.

> I brought over all of the tools from the monero_tools guide as well as the translations that were available for them

Usually the language files shouldn't be touched, but in these case i think should be fine because i locked the translations on Weblate some days ago, so there shouldn't be conflicts.

> Should I remove the monero_tools weblate .po files? (_i18n/[lang]/resources/user-guides/weblate/monero_tools.po)

I'll take care of that, this PR is dense enough.

> moneronodes.i2p.xyz can't be reached.

Remove

> nestorgames.com I think better belongs somewhere on the Merchants page?

yep

> monerobase.com seems to not have been updated since Dec 2018.

monerobese used to collect the meeting logs. They stopped updating it long ago, but could be still useful in case somebody wants to dig on old logs. We could add it, but specifying it's there only as an archive,

> xmrnotify.com redirects to https://darknetpages.com/monero-guide/

remove

A couple more things:

- You use two different styles for the text. In some cases it's the website name + description and in some cases is only description. We have to use only one of the two.
- I'm thinking would be probably useful to have a short description about what info are included in each box. In any case we should add a short description at the top, below  "Tools"

steverusso:
> You use two different styles for the text. [...] We have to use only one of the two.

Good call (I just copied over what was originally there). I changed it to use the `site - description` style, but it didn't make sense to me for the Block Explorers or Address Generators categories? Imo, those descriptions are in their titles.

---

> I'm thinking [...] a short description [...] in each box. In any case we should add a short description at the top, below "Tools"

I think a description at the top would suffice, something along the lines of whats on the merchants page, such as:

> Below is a list of third-party tools for interacting with the Monero ecosystem. **These tools are not vetted by the Getmonero team, see the disclaimer at the bottom of this page.** If a tool no longer supports Monero or you would like a Monero tool to be listed, please open a GitHub issue and let us know

And then I'll also add the disclaimer at the bottom?

erciccione:
> Imo, those descriptions are in their titles.

I agree

> And then I'll also add the disclaimer at the bottom?

Sounds good

_i18n/en.yml
```diff
@@ -866,6 +867,37 @@ moneropedia:
...
     clsag: CLSAG
 
+tools:
+  intro1: Below is a list of third-party tools for interacting with the Monero ecosystem. <b>These tools are not vetted by the Getmonero team, see the disclaimer at the bottom of this page.</b> If a tool no longer supports Monero or you would like a Monero tool to be listed, please
+  intro2: open a GitHub issue and let us know
+  block-explorers: Block Explorers
+  xmrchain: xmrchain.net
+  moneroblocks: MoneroBlocks
+  exploremonero: Explore Monero
+  monerovision: MoneroVision
```
erciccione, line 877: These and other entries below don't need to be on the yaml file, since they are not to be translated.

_i18n/en.yml
```diff
@@ -866,6 +867,37 @@ moneropedia:
...
+
+  address-generators: Address Generators
+  moneroaddress: moneroaddress.org
+  xmrllcoins: xmr.llcoins.net
+
+  network: Network
+  monerohash-nodes: monerohsah.com - Monero node map
```
erciccione, line 884: Assign to this entry (`monerohash-nodes`) only the description (`Monero node map`) no need to have the entire string translated. The name of the website can be directly on index.md.Same for the other entries below.

erciccione (changes requested):
I would say that the structure is a good start, we can improve it later on. Some fixes are needed before we can merge it

resources/tools/index.md
```diff
@@ -0,0 +1,90 @@
...
+            </div>
+        </div>
+    </div>
+    <div class="row">
+        <div class="left half col-lg-6 col-md-6 col-sm-12 col-xs-12">
+            <div class="info-block">
+                <div class="row">
+                    <div class="col">
+                        <h2>{% t tools.misc %}</h2>
+                        <p><a href="http://xmrtests.llcoins.net/checktx.html">xmrtests.llcoins.net - {% t tools.llcoins-checktx %}</a></p>
```
erciccione, line 75: If they don't have a TLS certificate we don't list them. Please remove `xmrtests.llcoins.net` and its related strings.

resources/tools/index.md
```diff
@@ -0,0 +1,90 @@
...
+        </div>
+    </div>
+    <div class="row">
+        <div class="left half col-lg-6 col-md-6 col-sm-12 col-xs-12">
+            <div class="info-block">
+                <div class="row">
+                    <div class="col">
+                        <h2>{% t tools.network %}</h2>
+                        <p><a href="https://monerohash.com/nodes-distribution.html">monerohsah.com - {% t tools.monerohash-nodes %}</a></p>
+                        <p><a href="https://moneroblocks.info/stats">MoneroBlocks - {% t tools.moneroblocks-stats %}</a></p>
```
erciccione, line 48: This is now `https://localmonero.co/blocks/stats`, so i would change name and description. Instead of 'Network Statistics' i would put 'Explorer and Statistics' and i would use the URL: `https://localmonero.co/blocks/`

resources/tools/index.md
```diff
@@ -0,0 +1,90 @@
...
+                        <p><a href="https://nownodes.io/connect-xmr-node">NOWNodes - {% t tools.nownodes %}</a></p>
```
erciccione, line 49: I think this is just an API provider. I don't see any standalone block explorer on this page.

resources/tools/index.md
```diff
@@ -0,0 +1,90 @@
...
+        </div>
+    </div>
+    <div class="row">
+        <div class="left half col-lg-6 col-md-6 col-sm-12 col-xs-12">
+            <div class="info-block">
+                <div class="row">
+                    <div class="col">
+                        <h2>{% t tools.misc %}</h2>
+                        <p><a href="http://xmrtests.llcoins.net/checktx.html">xmrtests.llcoins.net - {% t tools.llcoins-checktx %}</a></p>
+                        <p><a href="https://ownbit.io">Ownbit - {% t tools.ownbit %}</a></p>
```
erciccione, line 76: They don't seem to offer neither multisig or cold wallet for Monero anymore. Better remove them entirely.

erciccione (changes requested):

resources/tools/index.md
steverusso: That's why I put it under 'Network.' Are you suggesting moving it to 'Miscellaneous'?

erciccione: Yeah, that's probably better.

erciccione:
Please squash your commits into 1.

resources/tools/index.md
steverusso: I just realized:
* did you mean just to update the description in `en.yml` to not include the 'Explorer' part?
* if we are going to move this one to 'Miscellaneous', I'll also move `getblock.io` to there as well since that's also a node provider?

erciccione:
ping @steverusso. Let me know if you need help with squashing.

resources/tools/index.md
steverusso: Friendly ping @erciccione :)
Just wanted clarification on this move

erciccione: > did you mean just to update the description in en.yml to not include the 'Explorer' part?

Yeah, the 'Explorer' part should be removed.

> if we are going to move this one to 'Miscellaneous', I'll also move getblock.io to there as well since that's also a node provider?

Right. I think makes more sense to leave them where they are then.

erciccione (approved):
Looks good. Thanks :)

PR #1676, opened 2021-06-16
minor seo improvements

erciccione:
- optimize cumulative layout shift in the home
- ensure text remains visible during webfont load


erciccione:
Disregard. The CSS changes break the "arrow-up" button. Will look into it.

PR #1696, opened 2021-07-03
downloads: revamp section 'Hardware Wallets'

erciccione:
- After #1542 there has been no news about Kastelo. I'm assuming development has stopped for good, so i'm removing it from the section.
- I trimmed the two images (ledger and trezor) to make consistent styiling easier
- Added two different versions of the section: one for mobile and one for desktop

**Desktop Before:**
deskbefore

**Desktop after:**
desk

**Mobile before:**
mobbefore

**Mobile after:**
mob


selsta:
Just writing "Open Source" without any information is potentially misleading, e.g. Ledger apps are open source. The firmware and hardware isn't.

erciccione:
@selsta what do think would be more clear? We could be more specific about what is open source and what is not, but that will make the section much more verbose. Maybe we could just write "partly" and then we specify?

I don't know much about hardware wallets so any suggestion is appreciated.

selsta:
I don't know, just don't like having these potentially misleading ✓ / X icons while the reality is more nuanced. Adding "partly" would be more accurate with a pop up explaining it, but don't know how well that works on mobile.

erciccione:
> I don't know, just don't like having these potentially misleading ✓ / X icons while the reality is more nuanced

I agree with that. I just wasn't aware of the fact that none of them is fully open source.

> "partly" would be more accurate with a pop up explaining it,

Better a note than a pop up IMO. We could do something like `partly*` and then `* firmware and hardware are not open source*

> don't know how well that works on mobile

It's two different layouts for mobile and desktop, so that's not a problem

selsta:
Sounds good.

So Ledger: Open source (monero) app, firmware and hardware closed source
Trezor: No apps, open source firmware and hardware as open as possible

erciccione:
@selsta what about this:

selsta:
@ph4r05 Is this correct? Do you have any wording recommendations?

rating89us:
Even though it's still in beta, I would add Feather Wallet in "Wallets supported" column in both Ledger and Trezor.

rating89us:
I don't think this "Open source" column is really useful/needed here, since users are not going to visit this page when they are deciding which hardware wallet they are going to buy.  It also gives the impression that we are not being neutral, and we are saying "open source = good" vs. "not open source = bad", when this should be a user decision.

Instead, I would add a third column "Minimum required software" after "Model(s)" column, where we would list the minimum version required for each wallet model and its release date. It is a useful information, and it's common to have users that have an outdated firmware. Also, it would be an "official" place to check what firmware version is necessary for each wallet model, being an useful resource if you are giving support to an user.

Example (not using real values):

Firmware v1.6.1 (release date: 2021-01-01)
Firmware v1.2.4-5 (release date: 2020-11-20)
Monero App v1.7.6 (release date: 2021-02-02)

Firmware v2.4.1 (release date: 2021-07-21)

selsta:
Overall I agree that the min firmware version is more useful than if it is open source or not.

erciccione:
>  It also gives the impression that we are not being neutral

We are absolutely not neutral in that aspect. The website makes clear distinctions when we list open source vs closed source tools and we always favour FOSS software (only open source software is listed in the downloads page for example)

> Overall I agree that the min firmware version is more useful than if it is open source or not.

If that's the preference and people prefer to know the minimal firmware version i'm ok with that, but i need this info for each hardware wallet. 

> Even though it's still in beta, I would add Feather Wallet in "Wallets supported" column in both Ledger and Trezor.

Ok. I will specify it's still in beta.

selsta:
Ledger:
- Firmware v1.6.1 or newer
- Monero app v1.7.6 or newer

Trezor:
- Firmware v2.3.4 or newer

rating89us:
> but i need this info for each hardware wallet.

A better idea is to add a simple link to a dedicated page that explains in more details how much "open source" each hardware wallet is. By doing this we can give more detailed explanation (for newbies and tech-savvy users), since we're not restrained by the limited space of Downloads page. We should also keep in mind that probably more hardware wallets will support Monero in the future, and explaining "open source details" of each hardware wallet probably won't fit in Downloads page.

> Firmware v1.6.1 or newer

I agree with this presentation. Release date will ocupy too much space and it's not really relevant.

erciccione:
> Ledger Nano S has not the same firmware version (last firmware version is 2.0.0, released 4 May 2021) of Ledger Nano X (last firmware version is 1.3.0, released 18 May 2021).

What are the minimal firmware versions (for Monero) for these two?

selsta:
Ledger Nano S firmware: 1.6.1 or newer
Ledger Nano X firmware: 1.2.4-4 or newer

erciccione:
Suggested changes added and screenshots updated.

_i18n/en.yml
```diff
@@ -253,6 +248,16 @@ downloads:
...
   sourcearchive: Source Code (archive)
+  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by two hardware wallets: Ledger and Trezor."
+  hwth: Hardware Wallet
+  hwthmodel: Model(s)
+  hwthsupported: Wallets supported
+  hwthfirm: Min. firmware version supported
+  hwnanos: "Nano S: 1.6.1"
+  hwnanox: "Nano X: 1.2.4-4"
+  hwtrezor: v2.3.4
+  hwnewer: or newer
```
selsta, line 259: "Min. firmware version supported" makes "or newer" obsolete.

_i18n/en.yml
```diff
@@ -253,6 +248,16 @@ downloads:
...
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by two hardware wallets: Ledger and Trezor."
+  hwth: Hardware Wallet
+  hwthmodel: Model(s)
+  hwthsupported: Wallets supported
+  hwthfirm: Min. firmware version supported
+  hwnanos: "Nano S: 1.6.1"
+  hwnanox: "Nano X: 1.2.4-4"
+  hwtrezor: v2.3.4
```
selsta, line 258:
```suggestion
  hwtrezor: 2.3.4
```

for consistency

_i18n/en.yml
erciccione: fixed

_i18n/en.yml
erciccione: fixed

selsta:
Theoretically app version for Ledger is missing but the CLI / GUI now displays a message if the app is outdated so not sure if we want to explicitly mention it on the website.

Screenshot 2021-07-14 at 09 27 04

erciccione:
IMO it's fine to be more informative. People can see if their HW is supported by just looking at the downloads page. I think that's useful.

rating89us:
You forgot to add Feather Wallet (beta)

rating89us:
> Theoretically app version for Ledger is missing but the CLI / GUI now displays a message if the app is outdated so not sure if we want to explicitly mention it on the website.

> IMO it's fine to be more informative. People can see if their HW is supported by just looking at the downloads page. I think that's useful.

Actually selsta wants to display only firmware version for Ledger, instead of firware version + (ledger) monero app version. I agree with omitting monero app version.

_i18n/en.yml
```diff
@@ -253,6 +248,15 @@ downloads:
   desktop: Desktop
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by two hardware wallets: Ledger and Trezor."
+  hwth: Hardware Wallet
+  hwthmodel: Model(s)
+  hwthsupported: Wallets supported
+  hwthfirm: Min. firmware version supported
```
rating89us, line 255:
```suggestion
  hwthfirm: Min. firmware version
```

_i18n/en.yml
```diff
@@ -253,6 +248,15 @@ downloads:
   desktop: Desktop
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by two hardware wallets: Ledger and Trezor."
+  hwth: Hardware Wallet
+  hwthmodel: Model(s)
+  hwthsupported: Wallets supported
```
rating89us, line 254:
```suggestion
  hwthsupported: Wallets
```

_i18n/en.yml
```diff
@@ -253,6 +248,15 @@ downloads:
   desktop: Desktop
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by two hardware wallets: Ledger and Trezor."
```
rating89us, line 251:
```suggestion
  hwdesc: "Hardware wallets are cryptographically secure devices that help keep your coins safe. Monero is currently supported by Ledger Nano S/X and Trezor Model T."
```

erciccione:
@rating89us Suggestions and Feather added.

rating89us:
Ledger and Trezor logos are being resized to a point that they become pixelated. Maybe we shouldn't even resize them.

image

image

rating89us:
I suggest adding the following links to firmware column:
- https://support.ledger.com/hc/en-us/articles/360013349800 (Nano X firmware)
- https://support.ledger.com/hc/en-us/articles/360002731113 (Nano S firmware)
- https://wiki.trezor.io/User_manual:Updating_the_Trezor_device_firmware (Trezor firmware)

I also prefer the links to Ledger and Trezor articles being placed on the text (on "Models that support Monero" column) instead of on the logos.

erciccione:
> Maybe we shouldn't even resize them.

Yeah, fixing that.

> I suggest adding the following links to firmware column:

I don't think adding these links fits our download page. Better not overload the users with info they don't need. If they want info about upgrading their firmware, they can explore the website themselves (we provide links).

> I also prefer the links to Ledger and Trezor articles being placed on the text

Suggestions for what to use? It's not a problem with Trezor. We can point to the Monero page, but what should we use for ledger? we currently point to https://support.ledger.com/hc/en-us/articles/360006352934-Monero-XMR- from the image, but doesn't make sense to have two links (nanox and nano s) pointing to the same url.

erciccione:
Pushed fix for images.

selsta (approved):

PR #1731, opened 2021-07-16
Tools: add links to monero.fail node list and node map

rating89us:


erciccione (changes requested):
I'm not sure if it makes sense to add two links to the same website but a different page. Better to add on single entry. Like:

Monero.fail (node list and map) - https://monero.fail/

selsta:
+1 for one link

selsta (approved):

erciccione (approved):

PR #1733, opened 2021-07-16
Tools: add link to MoneroLogs

rating89us:


erciccione (changes requested):
I think this will conflict with #1722, probably better to add this commit to that pull request instead of having a dedicated PR.

selsta (approved):

selsta:
@erciccione conflict how? 1722 changes different files

erciccione:
I linked to the wrong PR. I meant #1732, which changes the same file and almost the same line.

selsta:
Should be fine to just rebase after 1732 is merged.

erciccione:
Yeah, my point was to avoid the conflict and the need of rebase after the merge by putting the two commits in the same PR. Not a big deal, just makes maintainance easier.

erciccione (approved):

PR #1750, opened 2021-08-01
Downloads: update GUI/CLI title and intro text

rating89us:


_i18n/en.yml
```diff
@@ -209,7 +209,7 @@ downloads:
   pruning: Blockchain pruning
   pruning1: Not enough disk space? Just use @pruning to download only 1/3 of the blockchain
   langs: "<b>30+ languages</b> available"
-  cli_intro: The CLI wallet gives you the total control over your Monero node and funds. Highly customizable and includes various analysis tools, as well as an HTTP RPC and 0MQ interface.
+  cli_intro: An open-source command line interface (CLI) wallet developed by the Monero community, completely free to use, best suited for developers, intermediate, and advanced users. The Monero CLI wallet gives you total control over your node and funds. Highly customizable and includes various analysis tools, as well as an HTTP RPC and 0MQ interface.
```
erciccione, line 212: This phrase should go in its own yaml key.

erciccione (commented):
I'm not a big fan of this change to be honest. It's more a rewording than an update and will result in the description needing to be retranslated in all languages. I'm not sure it worth it, but If others think makes sense we can merge it.

rating89us:
- Most users don't know that GUI means graphical user interface and CLI means command line interface.
- The wallet names are "Monero GUI (wallet)" and "Monero CLI (wallet)" and not just "GUI wallet" or "CLI wallet".
- It's not stated anywhere in Downloads page that Monero GUI and Monero CLI are free open-source software developed by the community, that anyone can participate in the development (unlike commercial wallets).

_i18n/en.yml
```diff
@@ -209,7 +209,8 @@ downloads:
   pruning: Blockchain pruning
   pruning1: Not enough disk space? Just use @pruning to download only 1/3 of the blockchain
   langs: "<b>30+ languages</b> available"
-  cli_intro: The CLI wallet gives you the total control over your Monero node and funds. Highly customizable and includes various analysis tools, as well as an HTTP RPC and 0MQ interface.
+  cli_intro: An open-source command line interface (CLI) wallet developed by the Monero community, completely free to use, best suited for developers, intermediate, and advanced users.
+  cli_intro2: The Monero CLI wallet gives you total control over your node and funds. Highly customizable and includes various analysis tools, as well as an HTTP RPC and 0MQ interface.
```
erciccione, line 213: The old placeholder must keep its original name or all translations for that string will be lost.

```suggestion
  cli_intro2: An open-source command line interface (CLI) wallet developed by the Monero community, completely free to use, best suited for developers, intermediate, and advanced users.
  cli_intro: The Monero CLI wallet gives you total control over your node and funds. Highly customizable and includes various analysis tools, as well as an HTTP RPC and 0MQ interface.
```

erciccione (changes requested):

_i18n/en.yml
rating89us: The current code of this PR keeps translation of the first part (see below):
- First part (cli_intro): old spanish translation of cli_intro
- Second part (cli_intro2): previous english sentence (untranslated)

image

erciccione: That's true as long as nobody touches the translated string. To weblate it will be an entirely different string, when in reality it just got renamed. That will cause the loss of the translated string for all languages at the next push. The old placeholder must keep its original content + changes, or translators will have to retranslate strings that were already translated even if only a single word has changed (`Monero`) in the string contained by the old placeholder.

rating89us: Ok, so I'll rename the new string to cli_intro0 instead of cli_intro2, because it will be displayed before cli_intro.

PR #1759, opened 2021-08-04
Downloads: differentiate local sync x remote sync (lightweight) wallets

rating89us:
Changes:
- Separate mobile/light wallets in two categories:
  - Local sync wallets: don't share any key with a remote server.
  - Remote sync (lightweight) wallets: share the private view key with a third-party server.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -249,6 +249,10 @@ downloads:
   desktop: Desktop
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  localsync: Local synchronization
+  localsyncinfo: These wallets connect to @remote-nodes and scan the @blockchain on your device, without sharing any key with the nodes. They provide higher privacy, but require more time synchronizing, since missing @blocks will be scanned only when the wallet is opened.
+  remotesync: Remote synchronization (lightweight)
+  remotesyncinfo: These wallets share your private @view-key with a remote server, which continuously scans the blockchain looking for your transactions. They are faster to use, but your privacy can be lessened if you don't control the remote server.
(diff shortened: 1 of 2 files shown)


_i18n/en.yml
```diff
@@ -249,6 +249,10 @@ downloads:
   desktop: Desktop
   antivirus: Antivirus
   sourcearchive: Source Code (archive)
+  localsync: Local synchronization
+  localsyncinfo: These wallets scan the blockchain on your device and don't share any key with remote servers. They provide higher privacy, but require more time synchronizing.
```
erciccione, line 253:
```suggestion
  localsyncinfo: These wallets connect to @remote-nodes and scan the @blockchain on your device, without sharing any key with the nodes. They provide higher privacy, but require more time synchronizing, since missing @blocks will be scanned only when the wallet is opened.
```

_i18n/en.yml
```diff
@@ -249,6 +249,10 @@ downloads:
...
+  remotesync: Remote synchronization (lightweight)
+  remotesyncinfo: These wallets share your private view key with a remote server, which scans the blockchain for you. They are faster to use, but your privacy can potentially be compromised if you use a third-party server.
```
erciccione, line 255:
```suggestion
  remotesyncinfo: These wallets share your private @view-key (needed to see incoming transactions) with a remote server, which scans @blocks looking for your transactions. They are faster to use because they continuously scan the blockchain looking for transactions, but your privacy can be lessened if you don't control the remote server.
```

erciccione (commented):
I'm ok with the change but i think simpler terminology should be used for the two descriptions. I'm afraid people will confuse it with remote/local nodes. The clearer, the better. I proposed some changes.

_i18n/en.yml
rating89us: I added most of your suggestion, but made a shorter version.

erciccione (approved):

PR #1761, opened 2021-08-05
custom: hide down arrow in tooltips

rating89us:
This PR removes a down arrow at the bottom of every tooltip.
image


erciccione:
I think the problem is related to the fact that the tooltip shouldn't show as in your screenshot. Should look like this:
p.

Probably better to open an issue.

PR #1767, opened 2021-08-07
Reorganize user guides, create new categories

rating89us:


resources/user-guides/index.md
```diff
@@ -35,6 +30,32 @@ title: titles.userguides
...
         </div>
     </section>
+    <section class="container">
+        <div class="row">
+            <div class="left half no-pad-sm col-lg-6 col-md-6 col-sm-12 col-xs-12">
+                <div class="info-block">
+                    <div class="row">
+                        <div class="col">
+                            <h2>{% t user-guides.backup %}</h2>
+                            <p><a href="{{site.baseurl}}/resources/user-guides/Offline_Backup.html">{% t user-guides.offline-backup %}</a></p>
```
erciccione, line 40: Do we expect more back up guides to be added? Otherwise this section will stay with one entry for a very long time. Maybe better to squash it into "General"?

rating89us: Yes, I have in mind at least 2-3 more guides. I think this topic is important and should have it's own category.

erciccione (approved):

PR #1772, opened 2021-08-08
new FAQ: 'What is ASIC resistance? Why is it important?'

erciccione:


_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
   avuln: >
     The Monero community has created a series of videos called "Breaking Monero", where potential Monero vulnerabilities are explored and discussed. There are 14 videos, with each exploring a different subject. Check out <a href="https://www.monerooutreach.org/breaking-monero/">the videos on monerooutreach.org</a>.
   vulnspotify: Available on Spotify as podcast
+  qasicresistance: "What is ASIC resistance? Why is it important?"
```
rating89us, line 504:
```suggestion
  qasicresistance: "What is an ASIC-resistant mining algorithm? Why is it important?"
```

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
...
+  aasicresistance: ASICs are basically special computers created to do only one job, contrary to normal computers, which are made for multiple purposes. This characteristic makes ASICs very efficient for @mining.
+  aasicresistance1: The problem is that these devices are very expensive, consume a lot of electricity and can be afforded by few, wealthy people. This leads to few entities owning a big amount of the hashrate of the network, which is a serious threat to the security of the network itself. For example, if big ASIC operators collude and manage to gain the majority of the hashrate of the network, they could arbitrarily reject transactions.
+  aasicresistance2: >
+    Monero fixes this problem by being ASIC-resistant: it uses an algorithm (@randomx) that makes impossible for ASICs to mine on the network. Miners must use common consumer hardware, which forces them to compete fairly. The Monero network is currently protected by thousands of computers. This results in a network much harder to attack, no miner having significant advantage over other miners (they all use more or less the same hardware) and less energy consumption.
```
rating89us, line 508:
```suggestion
    Monero fixes this problem by implementing an ASIC-resistant proof-of-work algorithm: it uses @randomx, an algorithm that strongly reduces the efficiency of ASICs mining, disincentivizing the development of this kind of hardware. Miners are economically incentivized to use common consumer hardware. As a result, the Monero network is currently protected by thousands of computers mining its algorithm using “regular” computers equipped with CPUs. This results in a network much harder to attack, since no miner has significant advantage over other miners (they all use hardware with similiar efficiency).
```

erciccione: Thank you for your reviews. My issue is that you tend to suggest complete rewords more than fixes or suggestions in your reviews. I'm not a big fan of this type of review, because if something is wrong i'm happy to correct it and if something is not explained well i'm happy to add suggested clarifications. But i don't agree with rewording an entire paragraph according to the style or wording preference of another contributor, if it's a one time thing it's fine, but it seems to be the norm in your reviews.

So, i will add some clarification where needed but i won't change it completely.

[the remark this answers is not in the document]
erciccione: Will fix

_i18n/en.yml
erciccione: This suggestion slightly changes the meaning of the entry and adds unnecessary complexity. I prefer to keep the original question.

_i18n/en.yml
erciccione: I added part of your suggestions.

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
...
+  aasicresistance: ASICs are basically special computers created to do only one job, contrary to normal computers, which are made for multiple purposes. This characteristic makes ASICs very efficient for @mining.
+  aasicresistance1: The problem is that these devices are very expensive, consume a lot of electricity and can be afforded by few, wealthy people. This leads to few entities owning a big amount of the hashrate of the network, which is a serious threat to the security of the network itself. For example, if big ASIC operators collude and manage to gain the majority of the hashrate of the network, they could arbitrarily reject transactions.
+  aasicresistance2: >
+    Monero fixes this problem by being ASIC-resistant: it uses an algorithm (@randomx) that strongly reduces the efficiency of ASICs, making them not profitable. Miners must use common consumer hardware, which forces them to compete fairly. The Monero network is currently protected by thousands of miners using 'regular' computers. This results in a network much harder to attack, no miner having significant advantage over other miners (they all use more or less the same hardware) and less energy consumption.
```
selsta, line 508:
```suggestion
    Monero fixes this problem by being ASIC-resistant: it uses an algorithm (@randomx) that strongly reduces the efficiency of ASICs, making them not profitable. Miners can use common consumer hardware, which allows them to compete fairly. The Monero network is currently protected by thousands of miners using 'regular' computers. This results in a network much harder to attack, no miner having significant advantage over other miners (they all use more or less the same hardware).
```

I would be careful with claiming that RandomX has a lower energy consumption. In general ASICs are more efficient if you consider hashes / watt, but it doesn't matter as the network is protected by the overall amount of energy put into it, not the efficiency of the algorithm.

Also once an ASICs becomes unprofitable it are thrown away, which causes silicon waste. This isn't the case with CPU mining, thus it can be argued RandomX is better for the environment.

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
   avuln: >
     The Monero community has created a series of videos called "Breaking Monero", where potential Monero vulnerabilities are explored and discussed. There are 14 videos, with each exploring a different subject. Check out <a href="https://www.monerooutreach.org/breaking-monero/">the videos on monerooutreach.org</a>.
   vulnspotify: Available on Spotify as podcast
+  qasicresistance: "What is ASIC resistance? Why is it important?"
+  aasicresistance: ASICs are basically special computers created to do only one job, contrary to normal computers, which are made for multiple purposes. This characteristic makes ASICs very efficient for @mining.
```
selsta, line 505:
```suggestion
  aasicresistance: ASICs are basically special computers created to do only one job, contrary to normal computers, which are made for general purposes. This characteristic makes ASICs very efficient for @mining.
```

I think general purpose is the correct word here, as theoretically you can create ASICs that are for multiple purposes.

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
...
+  aasicresistance1: The problem is that these devices are very expensive, consume a lot of electricity and can be afforded by few, wealthy people. This leads to few entities owning a big amount of the hashrate of the network, which is a serious threat to the security of the network itself. For example, if big ASIC operators collude and manage to gain the majority of the hashrate of the network, they could arbitrarily reject transactions.
```
selsta, line 506: See my other comment in regards to energy consumption. Also I would remove the mention of "wealthy" people, it doesn't really fit in. Many miners aren't wealthy, they just hope they can recoup their costs.

The other problem with ASICs is that ASIC manufacturers often use their own devices to mine before selling them to the public, so by the time they are available to the public they are outdated and unprofitable.

_i18n/en.yml
rating89us: You don't need to feel any pressure to accept or reject my suggestions. I'm not trying to impose my ideas or attack your contribution in any form, I'm just writing some additional information that you may find useful. Feel free to pick what you find interesting and discard what you think is unimportant.

_i18n/en.yml
erciccione: > Many miners aren't wealthy, they just hope they can recoup their costs.

My assumption was that if you can afford miners that cost thousands of dollars each, you are probably wealthy. But i see your point.

_i18n/en.yml
erciccione: I think it's a fair claim for a few reasons:

- Countrary to ASIC-mined chains, people can mine by using spare processor power of their pc. This saves electrivity compared to a brand new and dedicated mining system
- An ASIC-mined chain will always comsume more energy than a ASIC-resistant one, because the amount of electrical  energy put in the ASIC-resistant chain will be always less. If every single person in the world decides to mine Bitcoin, they will all need at least 1 ASIC that consume several Kw/h, while Monero users can use 1 or several consumer-grade machines, that doesn't go even near the amount of electricity used by an asic.
- While it's true that the point is the amount of energy put in the system, RandomX severlys limits the amount of energy a single entity can put in the system, because it's complex and inefficient to build rigs on RandomX. This is not a factor for ASICs. A single entity can drain huge amounts of power.

selsta: > Why change "must" with "can" and "forces" with "allows"? I chose that specific wording to make clear that the point of randomx is to force people to compete fairly, not "allow" them.

ASICs for RandomX are possible and thus they can use common computer hardware but nothing forces them to do so. Claiming that they "must" use it is inaccurate.

erciccione: My point was more that at that point a ASIC is as efficient as consumer hardware, but you are right, that wording is incorrect.

Any comment about my answer about energy consumption?

selsta: > An ASIC-mined chain will always comsume more energy than a ASIC-resistant one, because the amount of electrical energy put in the ASIC-resistant chain will be always less. If every single person in the world decides to mine Bitcoin, they will all need at least 1 ASIC that consume several Kw/h, while Monero users can use 1 or several consumer-grade machines, that doesn't go even near the amount of electricity used by an asic.

I don't know how you can make absolute statements like that. You can argue similarly that almost everyone in the world has access to an computer while only a small amount have access to ASICs thus something like RandomX can consume more energy because more people are able to mine.

> While it's true that the point is the amount of energy put in the system, RandomX severlys limits the amount of energy a single entity can put in the system, because it's complex and inefficient to build rigs on RandomX. This is not a factor for ASICs. A single entity can drain huge amounts of power.

RandomX allows for botnets which also results in a single entity consuming a lot of power.

---------

I find RandomX to be better for the environment due to less silicon waste, but the claim that it uses less energy seems unfounded to me and also useless (network security is energy put into the system so less energy is less security).

erciccione: I will just remove it for the sake of getting this merged.

erciccione:
Added all suggestions.

selsta (approved):

selsta:
Sorry, I didn't want to have a long discussion around this :P Approved it.

erciccione:
Don't be sorry, i appreciate your honest reviews :)

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
   avuln: >
     The Monero community has created a series of videos called "Breaking Monero", where potential Monero vulnerabilities are explored and discussed. There are 14 videos, with each exploring a different subject. Check out <a href="https://www.monerooutreach.org/breaking-monero/">the videos on monerooutreach.org</a>.
   vulnspotify: Available on Spotify as podcast
+  qasicresistance: "What is ASIC resistance? Why is it important?"
+  aasicresistance: ASICs are basically special computers created to do only one job, contrary to normal computers, which are made for general purpose. This characteristic makes ASICs very efficient for @mining.
+  aasicresistance1: The problem is that these devices are very expensive, consume a lot of electricity and can be afforded by few. This leads to few entities owning a big amount of the hashrate of the network, which is a serious threat to the security of the network itself. For example, if big ASIC operators collude and manage to gain the majority of the hashrate of the network, they could arbitrarily reject transactions.
```
luigi1111, line 506: I don't think "consume a lot of electricity" fits as a problem. The general issue revolves around access from my view. Cost is part of that, especially on the production side. Another is insider preference/self-mining incentive.

_i18n/en.yml
```diff
@@ -501,6 +501,11 @@ faq:
...
+  aasicresistance2: >
+    Monero fixes this problem by being ASIC-resistant: it uses an algorithm (@randomx) that strongly reduces the efficiency of ASICs, making them not profitable. Miners can use common consumer hardware, which allows them to compete fairly. The Monero network is currently protected by thousands of miners using 'regular' computers. This results in a network much harder to attack, no miner having significant advantage over other miners (they all use more or less the same hardware).
```
luigi1111, line 508: I think this should clarify that they are not profitable to build (at least hopefully). They certainly will be more profitable to mine with than gen purprose stuff.

_i18n/en.yml
erciccione: Removed.

selsta (approved):

Issue #1754, opened 2021-08-03
Discuss the possibility of migrating to a documentation platform


erciccione:
I've been thinking again about the possibility of moving the present documentation (developer guides and user guides) to a dedicated documentation platform (like [Readthedocs](https://readthedocs.org/)). This possibility came up from time to time during the years, but nothing ever came out of it. I think it worth having the discussion to see what community and devs think about it.

Keep in mind that such migration would take a significant effort. I cannot provide more details about that because i would have to look into the details of the platforms, but i want to see what people think about it first.

Pros:

- Developers are used to docs platforms. That would make easier for them to browse and contribute to the docs
- Possibility to have docs specific for each version
- Many documentation platforms have built-in multi-language support (managing translations would be easier)

Cons:

- The migration would take **a lot** of work (but depends by what we would want to include)
- Ideally we would need to register a subdomain like `docs.getmonero.org` where the docs would be hosted.

ProkhorZ:
For context, here's an example of a developer-focused Read the Docs site: [Solidity](https://docs.soliditylang.org/en/v0.8.7/#), and a user-focused Read the Docs site: [Finance Vote](https://financevote.readthedocs.io/en/latest/index.html). This particular platform is well-structured. On the other hand, the layout on the desktop is extremely vertical, on the left instead of the usual middle. It only takes up 40% of my screen, a third of which is used for the navigation pane.

I'm skeptical about this plan, because most of the documentation on this website is targeting users rather than developers. It could be useful to compile all documentation in one place, without removing the user guides. But that would be a big effort, and then it would have to be updated regularly for new versions. The examples I linked to were funded indirectly by token sales. There's a danger that outdated documentation would be copied blindly when there's no time to check everything.

In the past, the user guides had inconsistent formatting, but they look fine to me now.

ProkhorZ:
Side note: Is there a reason why there's no link to Mastering Monero on the Developer Guides page?

erciccione:
> I'm skeptical about this plan, because most of the documentation on this website is targeting users rather than developers

The dev guides are much more extensive than the user guides and are meant to be expanded.

> Also, keeping documentation for older versions available is not as relevant for Monero

This is not the case. For example, older versions might have missing RPC calls, or same calls but slightly different responses.

> Side note: Is there a reason why there's no link to Mastering Monero on the Developer Guides page?

I don't think Mastering Monero fits in the dev guides. We list it on the "Library" section.

[...]

PR #1778, opened 2021-08-13
moneropedia-mnemonic seed: different implementations, hints on storage

rating89us:


erciccione (commented):
The Moneropedia doesn't have a localization system, so changes should be done for all languages. If a paragraph is changed in the english version, should be changed in the translation as well, but translated blocks that are untouched in the english version should be left as they are. Things get complex when only minor changes are added to a paragraph, like in some cases here.

I know it's annoying, but as long as we don't have a system for the Moneropedia this is the situation. I'm working on solutions (one could be https://github.com/monero-project/monero-site/issues/1754), but in any case, it will take time.

erciccione (approved):
IMO we can merge it as it is, but maybe others want to review as well first.

plowsof:
polyseed now 16 words https://github.com/tevador/polyseed

_i18n/en/resources/moneropedia/mnemonicseed.md
```diff
@@ -1,19 +1,24 @@
...
+A 13, 14 or 25 word phrase used to backup a Monero @wallet, available in a number of languages. This phrase contains all the information needed to backup and restore a wallet. You should never share it, because anyone that has access to it will be able to view and spend funds from the wallet.
 
-### In-depth Information
+### How to store the mnemonic seed safely
 
-In the official wallet, the mnemonic seed comprises 25 words with the last word being used as a checksum. Those words correspond to a 256-bit integer, which is the account's *private* @spend-key. The *private* @view-key is derived by hashing the private spend key with Keccak-256, producing a second 256-bit integer. The corresponding *public* keys are then derived from the private keys.
+The most recommended way to store your mnemonic seed is to write two copies of it on a piece of paper, using a pen. Never store your mnemonic seed on a computer or any internet connected device. Don't store it on USB drives, email or cloud drives. Don't make screenshots or take photos.
```
plowsof, line 13: i'd go with the saying "if it's not in three places it doesn't exist", seed offset is also a possibility,, i wouldnt recommend multisig at the moment due to the current UX but we have RINO that could be mentioned,,  touch on some methods resistant against water/fire e.g. stamping them into washers secured between a nut and bolt

nahuhh: I don't like the idea of telling people how to store their seed. Also not sure why this is being changed, we already have a page on secure storage


writing on paper with pen sounds great, until 
1. someone finds your non-encrypted, plain text piece of paper and takes a picture
2. It is exposed to fire / water / dog ate my homework / lost in storage / found by your evil maid
3. someone produces a rubbing from the next page on the notepad
4. you get keylogged while typing it in to restore
5. etcetc


i dont understand why people think its safe. 
much easier to steal or make a copy of paper than to decrypt an encrypted backup.

personally, i think using a password manager with or without an airgapped device is infinitely more secure than pen and paper. add a never-stored seed offset for redundacy. 

https://www.getmonero.org/resources/user-guides/securely_purchase.html

```
save it as a PDF or text on a USB stick, burn it to CD/DVD, etc. Most likely you will want at least two or three copies, stored securely in different locations. If storing digitally, encrypt everything with a strong password. If storing on paper, do not show the wallet to anyone else who can memorize your 25 word key, or take a picture of the wallet without your permission. Sending someone a picture of the wallet is the same as giving away all of your funds.

Whichever method you chose, be sure there's no copy of the Monero wallet left over on the device you used. You may need to securely delete the Monero wallet if you saved it to a disk, or make sure your printer does not save a copy in memory.
```

plowsof: Agree on not providing advice here in such a brief paragraph (that contradicts the advice elsewhere).  The excerpt from securely purchase is sane. So a nack on hints of storage in this article.. its just for the different seed implementations

PR #1779, opened 2021-08-14
custom.css: increase bottom margin of fixed page header

rating89us:
This PR will fix overlaping of page anchors in user guides (on mobile) and will increase the bottom margin on downloads page (on mobile).


erciccione (commented):
This seems to cause inconsistent spacing above headers:
SP

rating89us:
You're right. I will close this now since I can't reproduce anymore the bug  I was getting

PR #1781, opened 2021-08-15
resources: update "How to restore a wallet from mnemonic seed" user guide 

rating89us:


rating89us:
Ready for review

_i18n/en/resources/user-guides/restore_account.md
```diff
@@ -1,53 +1,131 @@
...
+- You want to create a copy of your wallet on another device.
 
-- At the command prompt type:  `monero-wallet-cli --restore-deterministic-wallet`
+This guide will teach you how to restore a Monero wallet from its @mnemonic-seed.
 
-- Once you press enter you will be prompted for a wallet file name.  Give your wallet a new name, any name will do
+If you want to restore a wallet from its private keys, please read [this guide]({{site.baseurl}}/resources/user-guides/restore_from_keys.html).
 
-- Press enter again and you will be prompted for a password.  Give your wallet a new and long password
+If you want to restore a wallet from a hardware wallet, please check [this page]({{site.baseurl}}/downloads/#hardware).
```
erciccione, line 21: This page doesn't give information about restoring a hardware wallet. I would clarify or remove it altogether.

_i18n/en/resources/user-guides/restore_account.md
```diff
@@ -1,53 +1,131 @@
...
 
-- Press enter again and you will be prompted for the 25 word electrum style mnemonic seed that you retrieved earlier
+In order to restore a wallet from mnemonic seed, you must have a valid Monero mnemonic seed. A Monero mnemonic seed is a phrase with 13, 14, or 25 random words (multiple languages are supported). There are three types of Monero mnemonic seed:
 
--  You will then be prompted with "Restore from specific blockchain height (optional, default 0):" Default will start the restore process from the beginning of the Monero blockchain. If you don't know the specific blockchain height, just hit enter. (Specifying a specific blockchain height will start the restore process from that specific height. This will save a bit of time in scanning, if you know what starting blockchain height your initial funds were transacted for this specific account.)
+- **25 words mnemonic seed**: used by most Monero wallets and supported by all Monero wallets.
+- **14 words mnemonic seed with embedded restore height**: used and supported by Feather wallet.
+- **13 words mnemonic seed**: deprecated, previously used by MyMonero wallet. Supported by most Monero wallets.
```
erciccione, line 29: This could be much shorter. IMO it's not necessary to add a dedicated explanation of the various type of seeds. You already mention above that there are more than one.

_i18n/en/resources/user-guides/restore_account.md
```diff
@@ -1,53 +1,131 @@
...
+Alternatively, if you have access to the wallet being restored, you can find your mnemonic seed in the following locations:
 
-![choose language](/img/resources/user-guides/en/restore_account/choose-language.png)
+- **Monero GUI**: go to `Settings` > `Wallet` and open `Show seed & keys`.
+- **Monero CLI**: use the command `seed`.
+- **Cake Wallet**: open `Show seed/keys` on the main menu.
+- **Feather**: on the top menu click on `Wallet` > `Seed`.
+- **Monerujo**: open `Show Secrets!` on the main menu.
+- **MyMonero**: open your wallet and click on `Address`. Your mnemonic seed is shown as `Secret Mnemonic`.
+- **Edge**: long press the XMR wallet on the main page and open `Master Private Key`.
```
erciccione, line 47: I don't think we should add instructions for so many wallets. Only the GUI and CLI will do it. We don't need to provide instructions for every wallet we know. These guide will also need to be maintained.

_i18n/en/resources/user-guides/restore_account.md
```diff
@@ -1,53 +1,131 @@
...
+
+If you set a restore height lower than the block height of your first incoming transaction, Monero GUI will scan unnecessary blocks. In consequence, your initial wallet synchronization will take longer than necessary, but Monero GUI will still work.
+
+Therefore, if you're unsure about the block height of your first incoming transaction, it's always better to set a lower restore height instead of a higher one.
+
+![restore-height](/img/resources/user-guides/en/restore_account/restore-height.png)
+
+After entering your restore height, click on `Next` button.
+
+9) On the next page, give your wallet a strong password and confirm it. Since your wallet will hold your private keys and seed, it is really important to enter a strong password. The more you mix up letters (upper-case and lower-case), numbers, and symbols, the more secure your password is, and the harder it is for a brute force attack to discover it. It is NOT recommended to use a blank password or weak passwords like "password123" or "abcd1234".
```
erciccione, line 87:
```suggestion
9) On the next page, give your wallet a strong password and confirm it. Since your wallet will hold your private keys and seed, it is really important to enter a strong password.
```

It's not in the scope of this guide to give suggestion about password creation. Better leave that to the app themselves.

_i18n/en/resources/user-guides/restore_account.md
```diff
@@ -1,53 +1,131 @@
...
 
 ![wallet password](/img/resources/user-guides/en/restore_account/wallet-password.png)
 
-Specify your @daemon settings and click the `Right` arrow:
+10) If your wallet is in `Advanced mode`, the `@Daemon settings` page will be displayed. Here you can choose whether you want to start a local node or connect to a remote node. Choose your desired option and click on `Next` button.
```
erciccione, line 91:
```suggestion
10) If your wallet is in `Advanced mode`, the `Daemon settings` page will be displayed. Here you can choose whether you want to start a local node or connect to a remote node. Choose your desired option and click on `Next` button.
```

Don't use `@` if the text is between backticks

erciccione (changes requested):
I think this guide could be shorter. It would also be better if the original structure could be preserved. We have multiple translations for this guide and will be a waste to lose them because styiling changes.

Personally, where possible, i would prefer to keep the structure as it is (maybe keeping the additional 'When should I restore a wallet?') to avoid completely disregarding the translations we have, some of which were made by CCS funded contributors and funded by the community not long ago (for example German and Dutch).

Also, some sections that include more than 1 paragraph and images are duplicated for CLI and GUI. Would be better to have structure that include both, so that the text is not unnecessarely longer

_i18n/en/resources/user-guides/restore_account.md
rating89us: My original idea is to write a user guide on how to restore a wallet from a hardware wallet, but this guide doesn't exist yet. 

I could add a link to this non-existent page now and then we only merge this PR after the "how to restore a wallet from a hardware wallet" is created, or
I could remove it and then we add this link later, when the guide is created.

rating89us:
> I think this guide could be shorter.

I agree.

> It would also be better if the original structure could be preserved. We have multiple translations for this guide and will be a waste to lose them because styiling changes.
> 
> Personally, where possible, i would prefer to keep the structure as it is (maybe keeping the additional 'When should I restore a wallet?') to avoid completely disregarding the translations we have, some of which were made by CCS funded contributors and funded by the community not long ago (for example German and Dutch).

Current guide is really low-quality and with outdated screenshots. I don't think it needs to be preserved.

> Also, some sections that include more than 1 paragraph and images are duplicated for CLI and GUI. Would be better to have structure that include both, so that the text is not unnecessarely longer

Created a dedicated section to explain restore height in details.

_i18n/en/resources/user-guides/restore_account.md
erciccione: We cannot merge it with a link to a document that doesn't exist yet. If that's the case, let's change this PR into a draft.

rating89us: Ok, I removed the link.

rating89us:
Ready to be reviewed

erciccione:
I forgot to mention before that you don't have to update every single translation manually. We have po4a which does that work for us (we will have to run it anyway). See https://github.com/monero-project/monero-site/blob/master/po/po-update.sh. The config file will be added after the changes to the english file are merged.

rating89us:
Ok, I left only the English version changes.

erciccione (approved):

PR #1786, opened 2021-08-17
New FAQ: Is Monero for criminals?

### LINKED ISSUE #1782: Add new FAQ: "Is Monero used by criminals?"

erciccione:
Closes #1782

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -501,6 +501,11 @@ faq:
   avuln: >
     The Monero community has created a series of videos called "Breaking Monero", where potential Monero vulnerabilities are explored and discussed. There are 14 videos, with each exploring a different subject. Check out <a href="https://www.monerooutreach.org/breaking-monero/">the videos on monerooutreach.org</a>.
   vulnspotify: Available on Spotify as podcast
+  qcriminals: Is Monero for criminals?
+  acriminals: No, Monero is for everyone who values privacy, and not everyone who values privacy is a criminal. Monero is a permissionless, peer-to-peer, decentralized currency. This means anyone can participate on the network, and there is not a central authority to control how it is used and who uses the technology. Monero is a neutral tool in the hands of the public, very much like a kitchen knife, which can be used to slice a loaf of bread or stab an innocent person, or the US dollar, which can be used to donate do charity projects or fund terrorist attacks.
+  acriminals1: >
+    Monero, like every other currency, is money and can be used with good intentions (like funding charities) or bad intentions (like criminal activities). We cannot decide who uses Monero, but that doesn't mean we support criminal activities. Exactly like the Dollar is not considered 'for criminals', even if it's the preferred currency of terrorists, drug cartels and countless other criminal organizations.
+  acriminals2: Banning Monero would be a minor annoyance for criminals, but a major problem for those who use this currency to escape war, mass surveillance or sexual/political discrimination by tyrannical regimes or organizations.
 
 mining:
   intro1: Monero is a cryptocurrency that relies on proof-of-work mining to achieve distributed consensus. Below you'll find some information and resources on how to begin mining.
--- a/get-started/faq/index.md
+++ b/get-started/faq/index.md
@@ -34,6 +34,7 @@ permalink: /get-started/faq/index.html
                         <li><a href="#monero-meaning">{% t faq.qmoneromeaning %}</a></li>
                         <li><a href="#hardforks">{% t faq.qhf %}</a></li>
                         <li><a href="#videos">{% t faq.qvideos %}</a></li>
+                        <li><a href="#criminals">{% t faq.qcriminals %}</a></li>
                     </ul>
                     <li class="category">{% t faq.advanced %}</li>
                     <ul class="logo">
@@ -197,6 +198,14 @@ permalink: /get-started/faq/index.html
                             <li><a href="{{ site.baseurl_root }}/media/Monero%20-%20Ring%20Signatures.m4v">{% t faq.video_ringsig %}</a> - {% t faq.mvideos %} @Ring-Signatures</li>
                             <li><a href="{{ site.baseurl_root }}/media/Monero%20-%20RingCT.m4v">{% t faq.video_ringct %}</a> - {% t faq.mvideos %} @RingCT</li>
                         </ol>
+                    </div>
+                </div>
+                <div class="tab" id="criminals">
+                    <h3><a class="anchor" href="#criminals"></a>{% t faq.qcriminals %}</h3>
+                    <div class="tab-answer">
+                        <p>{% t faq.acriminals %}</p>
+                        <p>{% t faq.acriminals1 %}</p>
+                        <p>{% t faq.acriminals2 %}</p>
                     </div>
                 </div>
                 <div class="row center-xs">


_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
   avuln: >
     The Monero community has created a series of videos called "Breaking Monero", where potential Monero vulnerabilities are explored and discussed. There are 14 videos, with each exploring a different subject. Check out <a href="https://www.monerooutreach.org/breaking-monero/">the videos on monerooutreach.org</a>.
   vulnspotify: Available on Spotify as podcast
+  qcriminals: Is Monero for criminals?
```
rating89us, line 504:
```suggestion
  qcriminals: Was Monero created for criminals?
```

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  acriminals: No. Monero is permissionless, peer-to-peer and decentralized. This means Monero developers and community have no control over how is used and who uses the technology. Monero is a tool in the hands of the public, very much like a kitchen knife or the US dollar, both of which are currently used to commit or fund horrible crimes, like terrorist attacks.
```
rating89us, line 505:
```suggestion
  acriminals: No, Monero was created for everyone who values privacy, and not everyone who values privacy is a criminal. Monero is a permissionless, peer-to-peer and decentralized currency. This means anyone can participate on the network, and there is not a central authority to control how it is used and who uses the technology. Monero is a neutral tool in the hands of the public, very much like a kitchen knife, which can be used to slice a loaf of bread or stab an innocent person, or the US dollar, which can be used to donate do charity projects or fund terrorist attacks.
```

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  acriminals1: >
+    Monero, like every other currency, is money and can be used with good intentions (like funding charities) or bad intentions (like criminal activities). We cannot decide who uses Monero, but that doesn't mean we support criminal activities. Exactly like the Dollar is not considered 'for criminals', even if it's the preferred currency of terrorists, drug cartels and countless other criminal organizations.
+  qbanmonero: If you cannot avoid criminals using Monero, shouldn't it be banned?
```
rating89us, line 508:
```suggestion
  qbanmonero: If the Monero community cannot avoid criminals using Monero, shouldn't it be banned?
```

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  abanmonero: Banning Monero would be a minor annoyance for criminals, but a major problem for those who use this currency to escape war, mass surveillance or sexual/political discrimination by tyrannical regimes or organizations.
```
rating89us, line 509:
```suggestion
  abanmonero: A lot of crimes are committed over the internet, but no one advocates banning the internet. Banning Monero would be a minor annoyance for criminals, but a major problem for those who use this currency to escape war, mass surveillance or sexual/political discrimination by tyrannical regimes or organizations.
```

_i18n/en.yml
rating89us: Here it's more like: "Monero (like any other currency) is for everyone to use. It was not created specifically for criminals (like any other currency).

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  abanmonero1: Banning Monero is impossible by design, but even if it was possible, the only consequence of banning it would be the sudden rise of dozens of Monero clones. Exactly how it happened with Silk Road, the pioneer darknet market. Closing it resulted in countless more darknet markets being created, many of which didn't share the ethos of the original Silk Road
```
rating89us, line 510:
```suggestion
```

rating89us: I would remove this whole paragraph. Silk Road was a crime and Monero is not. Shouldn't be placed in the same category.

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  abanmonero2: "Another example is decades of war on drugs, still ongoing in most of the world. The result of this war is in front of everyone's eyes: countless of casualties and ruined lives, but drugs have never been so many and easier to find."
```
rating89us, line 511:
```suggestion
```
This one should also be removed. Monero community has no consensus about war on drugs.

_i18n/en.yml
```diff
@@ -501,6 +501,15 @@ faq:
...
+  abanmonero3: History teaches us that prohibition never succeeds, education does. If people want to do criminal activities, it's not the medium of exchange or weapon they use that makes the difference. If one is not available, they will use something else. It's the people who use it for legit reasons that will be negatively impacted.
```
rating89us, line 512:
```suggestion
  abanmonero3: History teaches us that most crimes can be solved through traditional investigative methods. There is no need to resort to banning a medium of exchange that is used by many legitimate users.
```

rating89us: I believe there will be no consensus in the Monero community to the affirmation that "prohibition never succeeds". I also wouldn't touch the guns topic here, since there will be no consensus about gun prohibition.

selsta (commented):
These are quite sensitive FAQ entries, I would prefer if we get multiple eyes on this. I agree with @rating89us suggestions.

_i18n/en.yml
erciccione: How is that related to my point? I'm saying that closing a darknet market will only result in many more being opened.

_i18n/en.yml
erciccione: This is not something the community needs to reach consensus on. It's history.

_i18n/en.yml
erciccione: This is not something the community needs to be in consensus about. The consensus we care about is the historian's one.

_i18n/en.yml
erciccione: I don't think this is useful, we already make multiple examples in the other paragraphs.

_i18n/en.yml
erciccione: That's not my point. My point is "Is Monero now the currency for criminals?". The point is not what it was created for, but how it's used.

erciccione:
I don't agree with most of the suggestions, which in some cases change the point of the entry. In some cases "consensus" in the community is requested for something that are just historical facts, which the scientific community agree on. I also find way overreaching that even just mentioning Silk Road makes people afraid of being compared to criminals when my original comparison is only about something being used for criminal purposes that will only come up again in different forms.

We don't need consensus for historical facts, but beside we didn't need consensus when we told people that Monero is for evading tyrannical regimes (which is very probably illegal), so i don't understand why it would be needed in such situation.

_i18n/en.yml
erciccione: This is a bit long, would be better to trim it, but if people want the change i'm ok with it.

_i18n/en.yml
erciccione: > I believe there will be no consensus in the Monero community to the affirmation that "prohibition never succeeds".

If that's somehow seen as a controversial statement, i'm ok with changing that phrase to something like "Prohibition is often not an efficient way to get rid of a problem. Education is"

> I also wouldn't touch the guns topic here

What do you mean? There is no gun topic here. I'm just saying that it's not the tool used to make a crime that makes the difference. We could remove it if people really want, but i would only remove that part. I'm ok with making changes if people want, but I don't agree with replacing my suggestion with yours, since it changes the paragraph completely.

erciccione: After a quick clarifying chat with selsta i agree the paragraph should be reworded, but not completely changed.

_i18n/en.yml
erciccione: If people are afraid of mentioning Silk Road for fear of association (i absolutely disagree with this), i guess we can avoid naming it or talk about it indirectly, but of course the impact of this paragraph on the reader will be heavily reduced.

rating89us:
The problem here is not whether your ideas are right or wrong. These topics (weapons, drug prohibition, drug markets) are highly controversial and shouldn't be part of an "official" position/opinion of Monero community on getmonero.org.

Also, we don't want to associate privacy with criminality/police matters. If you visit Tails OS or any VPN service website you will not read anything about criminals using it or it being banned (by governments). Imagine being a new user and reading these questions: "Is this VPN service for criminals?", "If you can't control who uses Tor, should TailsOS be banned?". It was this kind of association that was recently used against Monero on LWT (they joked on the fact that Monero promo video has even a police officer and a nightstick).

Here are some good examples of the kind of marketing we should follow:
image

image

Opinion content of controversial topics should be reserved for blogs, twitter, reddit or monero outreach.

So, on second though, I think this PR shouldn't be merged at all as getmonero.org FAQ, but it would be a good article for Monero Outreach.

erciccione:
> Also, we don't want to associate privacy with criminality/police matters

No we don't, i'm making these FAQ because the association monero-criminality already exists and it's actually very prominent and shadows the other aspects. If we ignore it and focus instead of saying why people need privacy (something we already say in detail in the entire website), the question "Is monero for criminals?" will stay unanswered.

> Also, we don't want to associate privacy with criminality/police matters.

VPNs and Tails are not primarely known for being tools for criminals like Monero is, it's time the community realizes that we need a proper and direct answer to these accusation. Going around it and say how Monero is great because protect your privacy is not helpful.

> These topics (weapons, drug prohibition, drug markets) are highly controversial and shouldn't be part of an "official" position/opinion of Monero community on getmonero.org.

This is very misleading. Nowhere in my text there is an official position/opinion about any of those three. The only clear reference is the failure of the war on drugs, which is a fact stated by the same people who started it, without mentioning the  "Global Commission on Drug Policy", which is an official organ. If there are personal opinions on weapons, drug prohibition and drug markets, i will be happy to remove them.

> Here are some good examples of the kind of marketing we should follow:

These are not FAQs, but landing pages. There is an abyss of difference with a FAQ.

selsta:
> The only clear reference is the failure of the war on drugs, which is a fact stated by the same people who started it, without mentioning the "Global Commission on Drug Policy", which is an official organ. If there are personal opinions on weapons, drug prohibition and drug markets, i will be happy to remove them.

That's not what I would call a fact. It's the opinion of "Global Commission on Drug Policy". The DEA will claim the opposite.

rating89us:
Why don't we just create a single question, more neutral:

`Who uses Monero? What is Monero used for?`

`Monero is used by users who values privacy, like activists, journalists, domestic violence survivors, people living on countries with repressive regimes. Monero is a permissionless, peer-to-peer, decentralized currency. This means anyone can participate on the network, and there is not a central authority to control who uses the technology and how it is used. Monero is a neutral tool in the hands of the public that can be used for good or bad things, very much like a kitchen knife, which can be used to slice a loaf of bread or stab an innocent person, or the US dollar, which can be used to donate to charity projects or fund terrorist attacks.`

selsta:
> Monero is a neutral tool in the hands of the public that can be used for good or bad things, very much like a kitchen knife, which can be used to slice a loaf of bread or stab an innocent person, or the US dollar, which can be used to donate to charity projects or fund terrorist attacks

I just worry that this plays straight into the hands of those that want to ban monero. "Look, they claim on the website you can use monero to fund terrorist attacks", even though if it would be disingenuous, they would quote it as a primary source.

erciccione:
> The DEA will claim the opposite and call it a fact.

Flat earthers will deny that the earth is round. That doesn't make it a non-fact. There will be always somebody/some entity denying a fact. The thing is that neutrality at all costs is not always the way to go and can be detrimental to the credibility of the project. Sometimes you need to call a spade a spade.

> Why don't we just create a single question, more neutral:

That question, even if could be useful, has a different scope then the original question "Is Monero used by criminals?"

As i said, i'm willing to remove what people think it's too controversial (even if i disagree and i think it will just be detrimental for the point we want to make), but if people don't want this entry or want to change its purpose i'll just close the PR.

rating89us:
Maybe change to
`Like any other currency, Monero is a neutral tool in the hands of the public that can be used for any activity that requires exchanging value between parties.`

selsta:
IMO we should take a look at other tools FAQ sections to get some ideas.

For example Tor has an abuse FAQ with a section "Doesn't Tor enable criminals to do bad things?": https://support.torproject.org/abuse/
Tor has user research on potential use cases: https://community.torproject.org/user-research/persona/

erciccione:
I removed the second entry ("If you cannot avoid criminals using Monero, shouldn't it be banned?") and moved one of its paragraphs to the "Is Monero for criminals?" question. I also added the suggested changes to one of the paragraphs.

> For example Tor has an abuse FAQ with a section "Doesn't Tor enable criminals to do bad things?"

That section is interesting, we should keep it in consideration for future FAQs.

rating89us:
Here a new suggestion partially based on your last update:

```
Q: Is Monero for criminals?

A: No, Monero is for everyone who values privacy, and not everyone who values privacy is a criminal. A large portion of Monero users are women survivors of domestic violence, citizens of third world countries trying to escape war and tyrannical regimes and citizens of first world countries trying to escape unlawful mass surveillance.

At the same time, Monero is a permissionless, peer-to-peer, decentralized currency. This means anyone can participate on the network, and there is not a central authority to control how it is used and who uses the technology. Like every other currency, Monero can be used with good intentions (like funding charities) or bad intentions (like criminal activities).

The Monero community cannot decide who uses Monero, but that doesn't mean we support criminal activities. We see Monero the same way everyone see the government money in physical form (cash or coins): a neutral tool in the hands of the public, that easily provides financial privacy to everyone.
```

erciccione:
> A large portion of Monero users are

As far as i know, we don't have any data that support of these claims.

selsta:
Yea, we shouldn't make claims like this without having the data. We can speak generally about the importance of privacy and use cases.

SamsungGalaxyPlayer:
I advise against having this as a FAQ

erciccione:
> I advise against having this as a FAQ

Could you be more detailed about why you think that?

We should address the fact that Monero is seen as a currency fro criminals by many.

erciccione:
Moving this to the merge queue.

SamsungGalaxyPlayer:
I think it's more likely to create a negative association that fix a negative association. The description also includes a crime, which is likely to have the opposite effect of breaking negative associations. Instead, it's likely that someone will read this and think "why is this FAQ here" and then have a takeaway from the page of "Monero" and "stabbing" together in the same sentence, even though the whole point is to try and debunk that narrative.

It's best to focus on the positives. Example FAQ questions:

- What civil rights groups use Monero?
- Why do civil rights groups use Monero?
- What free and open source (FOSS) projects use Monero?
- Why do FOSS projects use Monero?
- (etc)

SamsungGalaxyPlayer:
To emphasize this, I don't have a link, but Veep Season 3 Episode where the main character is trying to defame another character by saying "Danny Chung" and "torture" in the same sentences. Even though she is rebuking the allegations connecting these words together, all she is doing (purposefully) is further cementing the connection.

This is the best I have for clips:

https://getyarn.io/yarn-clip/da176537-5586-49fb-a3c5-9f3822b34799

erciccione:
> It's best to focus on the positives. Example FAQ questions:

These don't make the same point at all IMO, but people seem to agree so doesn't make any sense to spend more time on this. I'll just close this PR and the related issue.

PR #1810, opened 2021-09-03
README: fix introduction text

rating89us:
- Add a table of contents, like monero and monero-gui repositories use. 
- Remove "Introduction" title
- Some fixes to the introduction text

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![gh-actions/site](https://github.com/monero-project/monero-site/workflows/gh-actions/site/badge.svg?branch=master)
 
 ## Introduction
-This README here to walk you through everything you need to know to make changes, edits, or even completely new pages for the new [getmonero.org website](https://getmonero.org/). It'll definitely be a bit of a ride, so strap yourself in.
+This README will walk you through everything you need to know to make changes, edits, or even completely new pages for the [getmonero.org website](https://getmonero.org/). It'll definitely be a bit of a ride, so strap yourself in.
 Feel free to skip down to a relevant section if you already know what you need.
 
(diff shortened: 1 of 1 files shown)


erciccione (changes requested):
We removed the TOC not long ago (#1595) because it's provided by Github, so we don't need to maintain it. I don't see any benefit in adding it again.

I think substituting "README" with "guide" is not correct (maybe "document" would fit better), but in any case this a very minor and arbitrary change. I don't think worth having a dedicated commit for it. We should really limit changes related to personal preferences.

rating89us:
1) There is clearly an error in this sentence:
image

2) TOC was reintroduced in monero-gui README since TOC button is really hidden on github interface. Many users don't even know this button exists:
image
image

selsta:
> Not everyone knows what a "README" is. So I think it's better to use a more familiar term like "guide", "document", or "page".

Keeping it README should be fine, everyone capable of contributing to the monero-site repository will know what it is. Otherwise they won't read it anyway.

erciccione:
> There is clearly an error in this sentence:

Sure, let's fix it, but that's unrelated to change the word README to something else.

> TOC was reintroduced in monero-gui README

@rating89us I still think it's not necessary to add a TOC when the platform we use autogenerate one and it's just one click away. The fact that you preferred to re-add the TOC to the GUI and people had no problems with it it's not a good reason to add it again everywhere, especially when we removed it from this repo not long ago.

selsta:
Please squash.

selsta (approved):

erciccione (approved):

PR #1837, opened 2021-09-19
moneropedia: remove entry 'encryption'

erciccione:
It's basically just extracts from the Wikipedia entry 'encryption'. it's not really useful, so i would just remove it. It doesn't worth the effort to make it translatable, maintain it and translate it.

Also note:

- We never linked to this entry from inside getmonero (using `@`)
- If people write "encryption" on their search engine the first link they see is the wikipedia entry


selsta (approved):

QuickBASIC (approved):
Ack. This is a sensible change. It's superfluous.

PR #1846, opened 2021-09-24
Added SwapSwop

antonegorov:
Added SwapSwop to the list of Swappers resources


erciccione (changes requested):
NACK

This page only lists few and popular centralized exchanges and swappers.

PR #1848, opened 2021-09-25
developer-guides: wallet-rpc: fix validate_address example

### LINKED ISSUE #1849: developer-guides: wallet-rpc: validate_address example is incorrect
The example curl request for [validate_address](https://www.getmonero.org/resources/developer-guides/wallet-rpc.html#validate_address) should have the `params` set with `:` not `=`. Also, `allow_openalias` and `any_net_type` parameters are shown as strings and not boolean values. PR #1848 will fix.

reemuru:
Closes #1849 

The example for the validate_address wallet-rpc call is incorrect.
Parameter values for any_net_type and allow_openalias are shown as
string values instead of boolean. Also, the param object is
incorrectly being set with an equals sign but it should be a colon.

### DIFF
--- a/resources/developer-guides/wallet-rpc.md
+++ b/resources/developer-guides/wallet-rpc.md
@@ -377,7 +377,7 @@ Outputs:
 Example:
 
 ```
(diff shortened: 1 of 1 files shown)


erciccione (approved):
Thanks :)

reemuru:
@erciccione  Thanks for reviewing

PR #1859, opened 2021-10-03
delete 'node.' files that were generated by mistake

erciccione:
Generated by mistake while working on #1850


selsta (approved):

QuickBASIC (approved):
_I suggest we merge urgently because of the following._

The existence of these files breaks some tools on Windows as the Win32 API forbids the creation of files with a trailing `.` (My assumption is it's to retain backwards compatibility with the 8.3 filenames (which assumed that `file.` and `file` were the same.)

> Do not end a file or directory name with a space or a period. Although the underlying file system may support such names, the Windows shell and user interface does not. 

See: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#naming-conventions

Git on Windows is unable to sync these files and therefore assumes that they were deleted forcing contributors on Windows base their work on older versions of the site prior to the creation of these file or this PR.

PR #1894, opened 2021-12-03
Change press email to actual email for SEO reasons

SamsungGalaxyPlayer:
We want people to easily find this specific email address, and it's not like the existing situation is stopping bots anyway

### DIFF
--- a/press-kit/index.md
+++ b/press-kit/index.md
@@ -13,7 +13,7 @@ permalink: /press-kit/index.html
             <h2>{% t press-kit.contacts %}</h2>
         </div>
         <p>{% t press-kit.contactsp %}</p>
-        <p>{% t press-kit.contactgen %} press[at]getmonero[dot]org</p>
+        <p>{% t press-kit.contactgen %} press@getmonero.org</p>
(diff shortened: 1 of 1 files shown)


SamsungGalaxyPlayer:
We're getting more and more legitimate inquiries through this email, which is super good. Adding it to the press page made a huge difference.

erciccione (commented):
The important is that the way of showing email addresses is consistent. Note that initially email addresses on getmonero were obfuscated per a request of the core team, so we should see if they are fine with the change.

selsta (commented):
How would this improve SEO?

SamsungGalaxyPlayer:
@selsta it would only help slightly. It would also slightly help to generally make the website look more professional. It's a minor change overall, though I consider it a net positive.

erciccione:
As mentioned in my last comment: Email addresses are automatically hidden by the CDN, so this pr will result in the email being completely hidden instead. Closing. Feel free to open an issue if you want to suggest to not hide email addresses (assuming it's possible to disable the feature).

PR #1896, opened 2021-12-04
Rephrased awkward paragraph

ACK-J:
While reading through the site I came across a paragraph that felt clunky. I re-wrote it to be a little more clear. I'm not sure the proper way of going about fixing this for all other languages.


erciccione (approved):
Translations are dealt with automatically. In this case doesn't matter because the content doesn't really change. Thanks!

PR #1908, opened 2021-12-22
contributing: fix monero qr code and add alt text

erciccione:


fluffypony (commented):
Reviewed and confirmed the QR code is correct

PR #1913, opened 2022-01-07
Add Karsha to the Swappers section

khashmeshab:
Karsha was listed on the old version of this page, but somehow it got removed after the revamp. So I added it (us) back!

### DIFF
--- a/community/merchants/index.md
+++ b/community/merchants/index.md
@@ -104,6 +104,7 @@ permalink: /community/merchants/index.html
             <li><a href="https://simpleswap.io/">SimpleSwap</a></li>
             <li><a href="https://changenow.io/">ChangeNow</a></li>
             <li><a href="https://godex.io/">Godex</a></li>
+            <li><a href="https://karsha.biz/">Karsha</a></li>
           </ul>
       </div>
     </div>


selsta (commented):
Unfortunately we had a lot of problem with scam reports so we reduced the listed merchants, now we only list swappers that exist for a while and also have good reputation in the community.

I have googled karsha.biz and did not find much monero related feedback about it on Reddit. I think currently the swapper isn't ready yet to get added.

khashmeshab:
Well with all due respect, we were listed for years on the old version of this page. You may take a look at the history of the repository to verify my claim. We are around from 2009 and started exchanging Bitcoin since 2011. We are the oldest Bitcoin exchange. If we, with more than 12 years of history, aren't ready for the list, the who is?!

There are a few positive reviews on our TrustPilot page: https://www.trustpilot.com/review/karsha.biz. We are also listed on https://bitcoin.org/en/exchanges and https://www.bitcoincash.org/#exchanges. Also you can't find any negative review of us. You know, no negative review is a positive review itself! Since we respect anonymity, we attract a lot of Monero users. Around 60% of our orders are from/to Monero and are from tor. And you know tor users who use Monero? Asking them to write a review is too much to ask because they care too much about their privacy to write a review. God, they don't enter their email address to receive status updates of their orders.

We are definitely not a scam! Man, I "invented" the first Bitcoin web wallet! Please read more about us here: https://karsha.biz/about

selsta:
I didn't claim your website is a scam, I wrote that we had to remove all listings from the website because we received a lot of complaints about scammers in general, not your website specifically. The requirements to get added are more strict now.

If your website respect anonymity / allows Tor without KYC then try to advertise it on the r/monero subreddit to get some feedback.

khashmeshab:
I know that. I was just clearing things up. I admit we are not that good at advertisement, PR, and social networks. The thing is that we prefer to have our own loyal customers who found us through official websites like Bitcoin's and Monero's, rather than having lots of customers and law people see us and then having to deal with court orders and finally having to implement KYC and AML rules. Please look it my way: if we were on the Monero list of exchanges for years (I guess 5+ years) and you received no complaint about us and there's no negative review, that's a sign that we are a reliable partner, and we can keep our position in the list. Thank you.

erciccione (changes requested):
NACK

PR #1918, opened 2022-01-14
Add bottom margin for 404 page

HardenedSteel:
Before:
image
After:
image

### DIFF
--- a/404.html
+++ b/404.html
@@ -7,6 +7,6 @@ title: "Error 404, Page Not Found"
 	<div class="nopage">
 	    <h1 id="fourofour">404</h1>
 		<h1>{% t error.title %}</h1>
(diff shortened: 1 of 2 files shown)


HardenedSteel:
I broke somethings for deployment.

erciccione (changes requested):
Please do not use inline styling, but edit the `css/custom.css` file instead.

HardenedSteel:
Since all buttons have same class names i don't know how to change specific class without inline styling

erciccione (changes requested):
This is not an acceptable solution, as it would break other pages.

Could you share what browser are you using so we can test things out?

HardenedSteel:
@erciccione Firefox.

erciccione:
This PR cannot be accepted. I opened an issue to track the problem: #1920

Issue #1927, opened 2022-02-09
Listing Monetory.io on the Merchants&Exchanges page


monetory:
Hello  
I would like to suggest adding monetory.io to the [Merchants&Exchanges](https://www.getmonero.org/community/merchants/) page.
Monetory is an aggregator of p2p exchange offers. There are many independent exchanges that provide liquid Monero exchange and [monetory.io](http://monetory.io/) is nice way to find that offers. [Example](https://monetory.io/?request=1f57083fefe24b28beb3839de43ff01a).

Logo: 
![monetory_logo](https://user-images.githubusercontent.com/99361157/153286211-984360b2-a86d-4a1a-874f-15994cd732d5.svg)
Description: 
`The aggregator of P2P exchange offers. Collects offers from exchange platforms and independent traders.`
Open source: ❌
Non-custodial: `Depends by the seller`
No KYC (Know Your Customer): `Depends by the seller`
XMR <-> fiat: ✔
Onion address: ❌
link: `https://monetory.io`

SamsungGalaxyPlayer:
NACK; provides referral to other instant exchangers, for which there is already a place on the page

monetory:
> provides referral

project is an aggregator there's no fees for exchanges but there's referral codes
but we can turn off or change all referrals for traffic from getmonero.org

> other instant exchangers, for which there is already  

Can you provide an example? Can't find duplicate entries.

HardenedSteel:
I can add this to #2045 but just checked out XMR <=> fiat offers and most of them only in Russian or/and Russian payment methods which is unusable for most users. Other than Monero, most offers require KYC.

erciccione:
https://github.com/monero-project/monero-site/pull/1933#issuecomment-1611126292

PR #1933, opened 2022-03-07
Merchants: add monetory.io to merchants page

monetory:
Closes https://github.com/monero-project/monero-site/issues/1927


SamsungGalaxyPlayer:
NACK; provides referral to other instant exchangers, for which there is already a place on the page

monetory:
> provides referral

project is an aggregator there's no fees for exchanges but there's referral codes
but we can turn off or change all referrals for traffic from getmonero.org

> other instant exchangers, for which there is already  

Can you provide an example? Can't find duplicate entries.

erciccione:
PR got only negative feedback. More than a year has passed. Closing,

PR #1936, opened 2022-03-11
Update clearnet.md

Arsennikum:
Ошибка перевода. На англ "as you are still using clearnet (though you are more proxied than without a VPN)"


erciccione (changes requested):
Thanks for the PR, but changes to translations must be submitted on https://translate.getmonero.org/projects/getmonero/monero-site/ :)

PR #1943, opened 2022-04-13
Proposed fix for monero-dev Matrix links

baro77:
Today trying to enter in monero-dev Matrix room from Community > Workgroup, Element complained about missing room... checking the links and comparing them with other rooms' ones it seemed to me a couple of "#libera_" had been forgotten.
So I propose current PR fix ( I saw other links sometimes have postfixes, but I ignore the rationale behind them, so I just propose a vanilla link working for me today)


erciccione (approved):
Thanks! Those links are from before Libera got their own matrix server

PR #1957, opened 2022-04-30
Fix matrix link

Vyquos:
Add proper link to matrix room in Greek, English, Spanish, Polish and Russian translations.

### DIFF
--- a/_i18n/el.yml
+++ b/_i18n/el.yml
@@ -435,7 +435,7 @@ mining:
     επικίνδυνο
   hardware: Υλισμικό
   supportp: Αν έχετε ερωτήσεις ή απλά θέλετε να αντιπαρατεθείτε με άλλους εξορυκτές,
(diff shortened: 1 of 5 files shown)


erciccione (commented):
This could break things on Weblate. Would be better to submit the fixes directly there. Since it's a purely technical fix, maybe @netrik182 could take care of it directly on the platform

netrik182:
Just checked and there are plenty of other languages without 'matrix.to'. Including the source: https://translate.getmonero.org/translate/getmonero/monero-site/en/?checksum=3f53aecb670c7305

Let's not fix the translated ones here to avoid breaking things on weblate. Source en.yml should be addressed in this PR imho though. After source is updated I can manually fix all translations at once.

netrik182:
I fixed both source and translations at #1958. We can close this PR. Thanks @Vyquos for your suggestions.

Issue #1934, opened 2022-03-10
Please update or omit an outdated statement on the GetMonero.org website


LiveFree1776:
Hello,
On the website https://www.getmonero.org/get-started/what-is-monero/ there is an outdated statement:

"Unlike selectively transparent alternatives (e.g. Zcash), Monero is the only cryptocurrency where every user is anonymous by default."

Although it may have been true when written, Monero can no longer claim to be the "only" cryptocurrency that is anonymous by default. It may be the most trusted, time-tested, and peer-reviewed privacy crypto but it is no longer the "only" one. Could this be updated or changed by someone to reflect reality? Probably the simplest thing to do would be to change it to

"Unlike selectively transparent alternatives (e.g. Zcash), Monero is anonymous by default."

But it can be changed to anything as long as it is not a falsehood.

Thank you!

monerochan-moe:
he doesn't know what a pull request is and thinks a wagie is going to come here modify the html for him
![pepelaughz](https://user-images.githubusercontent.com/90231001/157584879-8e2e92f9-d61e-42a4-8ff6-5a9403699454.jpg)

LiveFree1776:
Of course I don't know what a "pull" request is. Who does? This literally is my first day on github. People told me if you want it to change make a request. So here I am. You can't say I didn't try to be the change I seek.

erciccione:
@monerochan-moe please refrain from trolling. People are allowed and welcome to open issues even if they don't know how to open pull requests. One thing doesn't exclude the other.

@LiveFree1776 I see your point. Maybe we could change it to something like "The only major cryptocurrency...". That would be more correct.

plowsof:
Please close, issue solved :+1:

PR #1961, opened 2022-05-07
Update confidential_para1 to "a" not "the only" cryptocurrency

citrus-flagman:
Changed "Monero is the only cryptocurrency where every user is anonymous by default." to "Monero is a cryptocurrency where every user is anonymous by default." to address issue [1934](https://github.com/monero-project/monero-site/issues/1934)


erciccione (changes requested):
As said in the issue, i think would be better to use `"The only major cryptocurrency" instead. A cryptocurrency is just... weak.

Please do not change the string in other languages. That needs to be done on weblate to avoid conflicts.

citrus-flagman:
@erciccione I have removed the incorrectly added translations and changed "a cryptocurrency" to "the only major cryptocurrency".

erciccione (changes requested):
The checks seem to be failing for some reason. Please squash the commit to only 1

citrus-flagman:
@erciccione I have squashed the commits to only 1 and it looks like the checks are passing now.

selsta (approved):

erciccione (approved):

PR #1969, opened 2022-05-23
FAQ in alphabetical order

plowsof:
The FAQ in the [get started](https://www.getmonero.org/get-started/faq/) tab is not in alphabetical order, so i made a [script](https://github.com/plowsof/userguide-drafts/blob/main/helper-scripts/add-a-faq/faq_alphabetic_order.py) to do that.

This would also be useful my other 'add a simple faq' automated [script](https://github.com/plowsof/userguide-drafts/blob/main/helper-scripts/add-a-faq/add_a_faq.py) that adds the new q/a in alphabetical order too.

### DIFF
--- a/get-started/faq/index.md
+++ b/get-started/faq/index.md
@@ -20,45 +20,45 @@ permalink: /get-started/faq/index.html
                 <ul>
                     <li class="category">{% t faq.general %}</li>
                     <ul class="logo">
-                        <li><a href="#anchor-word">{% t faq.qword %}</a></li>
+                        <li><a href="#vulnerabilities">{% t faq.qvuln %}</a></li>
+                        <li><a href="#videos">{% t faq.qvideos %}</a></li>
                         <li><a href="#anchor-contribute">{% t faq.qcontribute %}</a></li>
-                        <li><a href="#anchor-value">{% t faq.q1 %}</a></li>
                         <li><a href="#anchor-buy">{% t faq.q2 %}</a></li>
-                        <li><a href="#anchor-different">{% t faq.q4 %}</a></li>
+                        <li><a href="#anchor-value">{% t faq.q1 %}</a></li>
                         <li><a href="#anchor-btc-difference">{% t faq.q7 %}</a></li>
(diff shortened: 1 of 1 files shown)


selsta (approved):

erciccione (changes requested):
I don't think having the FAQ in alphabetical order would be beneficial. It's not a metric useful to the user and i think makes more sense to group the entries according to the content, but if people prefer it in alphabetical order we also need to make sure any future entry is added in the correct order, which is unlikely and we will probably end up having to fix the order from time to time.

That's my opinion, but if people prefer alphabetical order anyway and are willing to monitor long term that entries are inserted in the correct order, ok.

plowsof:
The [python script](https://github.com/plowsof/userguide-drafts/blob/main/helper-scripts/add-a-faq/faq_alphabetic_order.py) i made will take any FAQ page and order it alphabetically (hands free) (i am not that cruel to expect such manual labour ^^).

I have also made a 'Add a FAQ' [python script](https://github.com/plowsof/userguide-drafts/blob/main/helper-scripts/add-a-faq/add_a_faq.py) which will do alot of manual labour and order the new faq in alphabetical order.

ProkhorZ:
How would this affect translations? The script edits en.yml, which means that the translations would be sorted in the order of the English questions? That doesn't make sense.

I'd prefer to organize the FAQ better by using smaller, more specific categories than 'General' and 'Advanced'. E.g. Security, Privacy, History, Economics. 'Node and Wallet' can also be split in two, even though they're related.

erciccione (changes requested):
NACK for me. I don't see the benefit of adding complexity for the sake of having faq entries in alphabetical order in their own category. It just doesn't worth it. Beside, this approach would create issues that we currently don't have:

- Ordering faq entries in alphabetical order is not common practice. They are usually listed according to context.
- If the title of an entry changes, the entry could change position in the page
- Entries will be sorted in alphabetical order according to the English language as @ProkhorZ points out. So the change will be useless for all other languages (this effects #1973 as well)

PR #1967, opened 2022-05-17
user guides page alphabetic order

plowsof:
The tutorials state that alphabetic order is needed for the [user guides page](https://www.getmonero.org/resources/user-guides/). This was automated by [a script](https://github.com/plowsof/userguide-drafts/blob/main/helper-scripts/user-guide-index-alphabetic/index-alphabetic.py)  i made.

ba_1
ba_2
ba_3

PR #1973, opened 2022-06-01
Workflows: userguide index alphabetic test

plowsof:
Test will fail on push/pr of the user-guide index page if its not in alphabetic order / succeed if it is. will help repo maintainer with #1967 . (there is an easy-guide.py script to help with creating a guide / auto alphabetic ordering too, not in this PR) 

help for reviewer : im new to workflows so 'the names for jobs' may be wrong / vague 

To test, a PR can be made to [this repo](https://github.com/plowsof/userguide-drafts/blob/main/resources/user-guides/index.md) where the same script is being used (the index is currently in alphabetic order)

### DIFF
--- a/.github/alphabetic-userguide.py
+++ b/.github/alphabetic-userguide.py
@@ -1,0 +1,42 @@
+from bs4 import BeautifulSoup
+import yaml 
+import sys
+
+en_yaml = "_i18n/en.yml"
+index = "resources/user-guides/index.md"
+
+with open(en_yaml, 'r') as stream:
+    data_loaded = yaml.safe_load(stream)
+with open(index, "r") as f:
+    soup = BeautifulSoup(f, "html.parser")
+
+divs = soup.find_all("div", {"class": "col"})
+data = {}
+for div in divs:
+    category = str(div.find("h2"))
+    category = category.split(".")[1].split(" ")[0]
+    data[category] = []
+    items = div.find_all("p")
+    for raw_html in items:
+        title_key = raw_html.text.split(".")[1].split(" ")[0]
+        title_yaml = data_loaded["user-guides"][title_key]
+        info = { 
+                    "title_key": title_key,
+                    "title_yaml": title_yaml,
+                    "raw_data": raw_html
+                }
+        data[category].append(info)
+
+success = 1
+for category in data:
+    original = data[category]
+    alphabetic = sorted(data[category], key=lambda d: d['title_yaml']) 
+    data[category] = alphabetic
+    if original == alphabetic:
+        print(f"--> Category: {category} [ OK ]")
+    else:
+        print(f"--> Category: {category} [ FAIL ]")
+        success = 0
+
+if success == 0:
+    sys.exit(1)
--- a/.github/workflows/alphabetic-userguide.yaml
+++ b/.github/workflows/alphabetic-userguide.yaml
@@ -1,0 +1,22 @@
+---
+name: User-guides alphabetic ordering
+on:
+  push:
+    paths:
+      - 'resources/user-guides/index.md'
+  pull_request:
+    paths:
+      - 'resources/user-guides/index.md'
+jobs:
+  alphabetic-order:
+    name: User-guides alphabetic ordering
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt-get install -y --no-install-recommends python3-pip
+          sudo pip install beautifulsoup4 PyYAML
+      - name: Alphabetic-order script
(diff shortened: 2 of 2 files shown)


.github/workflows/alphabetic-userguide.py
```diff
@@ -0,0 +1,42 @@
...
+    soup = BeautifulSoup(f, "html.parser")
+
+divs = soup.find_all("div", {"class": "col"})
+data = {}
+for div in divs:
+    category = str(div.find("h2"))
+    category = category.split(".")[1].split(" ")[0]
+    data[category] = []
+    p = div.find_all("p")
+    for lol in p:
```
selsta, line 20: As this is something we might merge to the repo it would probably be better to use a proper variable name here :D

plowsof: Oops, thank you, my mind does go blank sometimes, adjusted it now, should be ok :+1:

selsta (commented):
One more suggestion, can you put the python script in the .github folder?

I think it's best if the workflows folder really only has workflows.

Similar to how it's done in the GUI: https://github.com/monero-project/monero-gui/tree/master/.github

selsta (approved):

erciccione (commented):
I appreciate the effort @plowsof, but i have a few concerns that make me doubt the benefits of this feature.

- Do we really need a script for this? During review is quite trivial to see if an entry is added in alphabetical order and if it's not you can just ask the contributor to change the order. It would be a 1 minute fix. This is the approach we used for the old merchants page and we never had problems with it.
- A python script is inconsistent with the other scripts we use and that integrate with jekyll. See the `_plugins` folder
- Few user guides are usually added to getmonero. If we are being very optimistic and assuming a drastic increase of new user guides (which is unlikely), this script + workflow will run maybe 2-3 times a year.
- I don't think it's needed (see point above), but all this could be replaced by ruby plugin that checks if the list is in alphabetical order before building the website, and automatically sort it if it's not.
- If the structure of the user guides page changes, this script will need to be adapted. So we will need somebody to do it. Not a big deal, but it's something else that needs to be maintained, especially since it's tied with the success or failure of tests.

plowsof:
Thanks for the detailed reply (more than few concerns :D )
>Do we really need a script for this? During review is quite trivial to see if an entry is added in alphabetical order and if it's not you can just ask the contributor to change the order. It would be a 1 minute fix. This is the approach we used for the old merchants page and we never had problems with it.

We don't need any checks, all problems can be fixed with time. 

>A python script is inconsistent with the other scripts we use and that integrate with jekyll. See the _plugins folder. If we are adding scripts, better stick with Ruby.

Breaking a build because of incorrectly ordering a page might be discouraging for a contributor, rather than pushing it to github and finding out then (where people can help / link them to a ['easy-guide'](https://github.com/plowsof/userguide-drafts/tree/main/helper-scripts/make-a-guide) script)

>Few user guides are usually added to getmonero (last addition was 9 months ago). If we are being very optimistic and assuming a drastic increase of new user guides (which is unlikely), this script + workflow will run maybe 2-3 times a year.

After reading the 'how to add a user-guide' docs, i understand why not many guides are added. With my script you just point it to a github URL. I then made 2 drafts for guides. [i2p zero gui with monero gui](https://github.com/plowsof/userguide-drafts/blob/main/i2p/monero-gui-i2p-node.md) and also [pruned node with monero gui](https://github.com/plowsof/userguide-drafts/blob/main/pruned-monero-gui.md). They are 1 click and ready for a PR.

The FAQ is another story, the docs literally say 'a guide would be too complicated'. Also have a 1 click deployment.

>I don't think it's needed (see point above), but all this could be replaced by a ruby plugin that checks if the list is in alphabetical order before building the website, and automatically sort it if it's not.

Interesting idea. Automatic sorting is the most ideal solution. But read your last and first point, an external check would be easier to maintain as you can just delete it from the workflow when things change / tell the contributor to fix it / use the script.

(i doubt im going to change your mind, but i just wanted to pay you the same respect of taking the time to comment)

selsta:
> Do we really need a script for this? During review is quite trivial to see if an entry is added in alphabetical order and if it's not you can just ask the contributor to change the order.

But why not automate it so that it can't be missed during a review process? It doesn't cost us anything and it doesn't clutter up the repo as it's all contained in `.github`. I don't see any downside.

> A python script is inconsistent with the other scripts we use and that integrate with jekyll.

Using python for this is sensible choice, we have something similar in the GUI repo even though we don't use python for the GUI. The script is part of a Github workflows process, not the website itself.

If this were something that would be executed during the website build process than I would agree with you, but that's not the case here.

erciccione:
Proposed to revert this pr in #2008

PR #1984, opened 2022-06-13
add moneropedia to po script

netrik182:
this will be later used to update .md, .pot and .po files of both moneropedia and user-guides.


erciccione (commented):
Probably better mention in the comment inside the script that it would work also for the Moneropedia

netrik182:
> Probably better mention in the comment inside the script that it would work also for the Moneropedia

Let me know if it's ok this way and I'll squash my commits.

erciccione (commented):
I'd say let's squash and then merge. @netrik182 hasn't been around for some time. If in a week or so the pr is stalling, i'll take care of recovering it.

erciccione:
recovered in #2142

PR #1986, opened 2022-06-21
A new Zero-Knowledge cheatsheet and related section name "make-up"

baro77:
Hi everybody

After the previous cheatsheets the intention was to write the next one about Bulletproof (the "to be continued..." of the third infographic). However very soon it has been clear that Zero Knowledge Proofs theory is mandatory to have a comprehensive understanding about how BP works (apart from being a very interesting field by itself, laying foundations for many cool techs): so I have authored and I'm PRing this new cheatsheet, as a step of an ongoing documentation effort.

I have tried to not miss the basic core concepts despite the simplified approach and topics selection, stressing ideas more than formalism: the merit of this cheatsheet, if any, is trying to represent an at-hand, concise, first bridge between theoretical foundations and applications developers approaching this world (not necessarily cryptography professionals). Of course this kind of resource cannot be a rigorous reference, but authoritative sources that I have consulted are listed, to let the interested reader dig deeper: definitely not an easy task, but worth the study these books require.

I have taken occasion for a small "rebranding" of cheatsheets section: as said I consider them as an ongoing documentation thread, in which boxes colours have a semantic meaning (even if weak) but also a unifying function as a recurring theme... and they really remind me salmons colours, from raw to cooked 😉  I hope the name I'm proposing is not too shallow, but if it is I'll not complain reverting to the current published one.

See ya


baro77:
Hi everybody

any news? I link the deploy preview to make more comfortable to take a look at the cheatsheet :)
https://deploy-preview-1986--barolo-time-757cf9.netlify.app/library/ZKbasicsCheatsheet20220621.pdf

baro77:
you are right @SamsungGalaxyPlayer , it wasn't my intention to mortgage the space :) ...just unconsciously beginning to think I was the only one with a cheatsheets' "fetish" lol 

I think "X cheatsheet by Salmon" could wrongly suggest "Salmon" is a nick of mines, so I guess a better option would be to rollback the section name as you suggest and to list my cheatsheets (with unchanged names/descriptions) as a bullet list in a group named "The Salmon Series". What do you think about it?

I'm going to wait a few days before re-commmiting and squashing, to gather any further opinions

baro77:
I have made the changes stemmed from the talk with @SamsungGalaxyPlayer and squashed to make the merge clean:
https://deploy-preview-1986--barolo-time-757cf9.netlify.app/library/

_i18n/en.yml
```diff
@@ -891,15 +891,15 @@ library:
...
+    <li><a href="RingsCheatsheet20210301.pdf">Rings Cheatsheet</a> - From generic legacy signature to CLSAG, a visual-holic journey through Rings flavours, presenting their core properties step-by-step in a poster-size infographic.</li>
+    <li><a href="RctCheatsheet20210604.pdf">RingCT Cheatsheet</a> - The recipe of a delicious RingCT Type 5 transaction: one CLSAG every eleven UTXOs, Moneroj amounts to taste; thicken everything with Pedersen Commitments whose overall equilibrium will be evident to waiters as well, and protect last ones with a BulletProof glaze; serve on chosen Stealth Addresses and consume not before ten blocks later. Cheers!</li>
+    <li><a href="ZKbasicsCheatsheet20220621.pdf">ZK Basics Cheatsheet</a> - A lightly theoretical interlude (hopefully still gentle, trying to stress concepts more than formalism and selecting the approached topics) to lay the foundations for Bulletproof and other future Zero-Knowledge-related features, if any.</li>
+    </ul>
```
erciccione, line 902: Better move the html tags to index.md as much as possible.

baro77:
Hi @erciccione 
nice to see you back here! 👍 
I got your point about html tags location.. it has been my doubt as well, but I followed Z2M and MM sections as tie-breakers (they both have links and bullet list in en.yml) because confining html tags to index.md would have implied a bunch of variables (instead of only two) interleaved with tags/links.  I'm just saying this because it seems relevant to me , anyway no problem to follow your suggestion if you confirm your preference :-)

erciccione (changes requested):
Adding variables to the html page is the correct approach. the LANG.yml files should be as clear of html as possible. This mostly to avoid that translators making a mistake will end up in a broken page. They are usually non technical people and it's better to give them only text that need to be translated. So, as a general rule, as much html should go where the rest of the html is (In this case `library/index.md`) and en.yml should be used only for strings. THe proposed change would also result in translators needing to translate again the same strings.

Better revert to the original structure and add the new cheatsheet according to that.

baro77:
@erciccione ok I got it... now that I understand the reason (thanks for the explanation!) perhaps I can get what you are looking for without reverting to the previous structure (I like the section style stemmed from talk with samsunggalaxyplayer)... please gimme a few days to try refactoring it before final decision

baro77:
Here it is, I think I have fully decoupled localize-able strings from html tags :)

erciccione (approved):
Looks good. Thanks @baro77 :)

PR #1990, opened 2022-06-29
Added 8 to make example more clear

### LINKED ISSUE #1947: Moneropedia address article clarification
In the [Moneropedia address article](https://www.getmonero.org/resources/moneropedia/address.html)  it says raw addresses start with a 4, and then it seemingly gives the monero donation address as an example, which does not start with a 4. Could this do with some clarification?

Mark-777-0:
Some exchanges have 8 to start their wallets: Binance, Bitfinex, KuCoinhave wallet adresses starting with digit '8'.

And the monero donate address starts with an 8, so saying they start with 4 might be confusing as alphatomcat said.

Closes #1947

### DIFF
--- a/_i18n/en/resources/moneropedia/address.md
+++ b/_i18n/en/resources/moneropedia/address.md
@@ -7,7 +7,7 @@ summary: "either an alias, such as donate.getmonero.org, or a set of 95 characte
 
 ### The Basics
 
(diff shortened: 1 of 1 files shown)


selsta (approved):

selsta:
@Mark-777-0 add Closes https://github.com/monero-project/monero-site/issues/1947 in your first comment so that it automatically closes, it doesn't work when you write it in the second comment.

Mark-777-0:
> @Mark-777-0 add Closes #1947 in your first comment so that it automatically closes, it doesn't work when you write it in the second comment.

I added it, but not sure if it worked.

erciccione (approved):

PR #1991, opened 2022-06-30
Fix typo

thecashewtrader:


thecashewtrader:
> I don't think this is a typo, since Moneroj is the plural of Monero

Oh my bad, should've looked it up.

PR #1992, opened 2022-06-30
Update HF blog post date/height

sethforprivacy:


SamsungGalaxyPlayer:
Should the filename change too?

sethforprivacy:
It would break all of the other links that have been shared if we do so, so I'd recommend against it unless we just want to put out an entirely new blog post in addition to this update.

sethforprivacy:
You can see the updated post/title here, @SamsungGalaxyPlayer: https://deploy-preview-1992--barolo-time-757cf9.netlify.app/2022/04/20/network-upgrade-july-2022.html

SamsungGalaxyPlayer (approved):

tobtoht (approved):

PR #1952, opened 2022-04-27
New sponsor

MajesticBank:
Adding new sponsor on site

PR #1994, opened 2022-07-08
add MajesticBank to sponsors

plowsof:
squashed version of #1952 full credits to @MajesticBank (with the 'Go the the CCS -> Go to the CCS' change in en,yml that recently got merged) (original PR seemed stagnant after squashing was requested, and since then, i introduced a small change to the `en.yml` file, which would have caused some headache for the submitter)


selsta (approved):

community/sponsorships/index.md
```diff
@@ -55,6 +55,18 @@ permalink: /community/sponsorships/index.html
                 </div>
             </div>
         </div>
+        <div class="row">   
+            <div class="left half no-pad-sm col-lg-6 col-md-6 col-sm-12 col-xs-12">
```
erciccione, line 59:
```suggestion
            <div class="right half no-pad-sm col-lg-6 col-md-6 col-sm-12 col-xs-12">
```

selsta (approved):

plowsof:
Thanks, added "left" -> "right" fix, and selsta spotted the paragraph id "tari" so this is now -> "mbank".  Are these id's used? only Tari has this

erciccione (approved):
@plowsof might be that we use the tari one (i don't remember)

PR #2005, opened 2022-07-26
Added Spendabit URL

Fxvargas:
This PR adds a link to Spendabit, which allows for filtering on merchants that accept XMR... For example, if you search for "phones", you can then select to filter on merchants that take Monero.

For full disclosure, I work with the team at Spendabit. But we hope this will be a valuable resource to the Monero community.


erciccione (commented):
This has been open for quite some time. If there is no feedback within a week or so, i'd close it

PR #2008, opened 2022-07-30
Revert "Workflows: userguide index alphabetic test"

erciccione:
This reverts commit 51ff4aa146d903701ac053e337823c44e8d0bcc1.

The commit (PR #1973) adds alphabetical order for user guides, but it works only for English and it's broken for all other languages. Reverting to fix unless there is a better option.


plowsof:
just following the rules in the readme "Add guide using markdown in the correct category, and in alphabetic order," which i added 
 as a feature of the automated create a user guide script here https://github.com/plowsof/userguide-drafts/tree/main/helper-scripts/make-a-guide

can you clarify what broken means? perhaps i can only have it trigger when the English version is modified if this is providing a false positive

please modify this PR to remove the alphabetic rule from the readme.

plowsof (changes requested):
lets remove/change this in the readme, in this, or a separate PR (as there are other things which are incorrect e.g. we don't do all languages anymore because of weblate)
```
Add guide using markdown in the correct category, and in alphabetic order, in ALL LANGUAGES to /_i18n/[ALL LANGUAGES]/resources/user-guides/index.md being careful not to mess with any indentation
```

plowsof (approved):

PR #2014, opened 2022-08-09
add Guarda

bchrvl:
A suggestion to add Guarda Wallet to the list of Light Wallets, they allow you to view the TX Key and the address import is free.


nahuhh:
Is guarda updated for v0.18 hard fork?

nahuhh:
> > 
> 
> Yes, Guarda is updated for v0.18.

@bchrvl 

Are you sure?
It appears that none of the Mymonero wallets are working.

bchrvl:
> > > 
> > 
> > 
> > Yes, Guarda is updated for v0.18.
> 
> @bchrvl
> 
> Are you sure? It appears that none of the Mymonero wallets are working.

I am sure that they have updated however there are problems after this update which they are currently trying to resolve.

bchrvl:
@nahuhh @erciccione, Would you be able to connect me to someone from the Monero team? We require some help with API. More precisely, looking for API docs describing methods, params/ parameter types, and the workflow (like the sequence of methods request). Also, where do we obtain the API keys to interact with the blockchain?

HardenedSteel:
> @nahuhh @erciccione, Would you be able to connect me to someone from the Monero team? We require some help with API. More precisely, looking for API docs describing methods, params/ parameter types, and the workflow (like the sequence of methods request). Also, where do we obtain the API keys to interact with the blockchain?

I don't know which API are you talking about but if you want to contact with team check [here](https://www.getmonero.org/community/hangouts/).

PR #2019, opened 2022-08-12
downloads: mark wallets that could leak data

### LINKED ISSUE #2011: More prominently display wallets that don't send any data by default in the download section
After the controversy risen in https://github.com/monero-project/monero-site/pull/2007, was pointed out that some wallets listed on getmonero might have a controversial privacy policy and/or ping home (sending sensitive data) without the user's specific consent. This can result in the privacy of the user being compromised.

Users of getmonero should be able to make an informed choice about what wallets listed on getmonero don't communicate externally unless specifically requested by the user.

We should show a badge or a note near the wallets that don't send data by default. We could also do the opposite, noting which wallets do send data.

To my knowledge, currently the wallets listed on getmonero that send data back home by default are:

- ~~Cake wallet~~ (edit: not anymore)
- Edge
- Mymonero

erciccione:
Closes #2011

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -286,6 +286,7 @@ downloads:
   hwnanox: "Nano X: 1.2.4-4"
   hwtrezor: "2.3.4"
   hwmobmodels: "Models that support Monero:"
+  noteleak: "* These wallets might ping their servers by default, potentially leaking the IP address of the user and other sensitive data."
 
 press-kit:
   intro: On this page you'll find press contacts, symbols and logos.
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -242,7 +242,7 @@ permalink: /downloads/index.html
                             <th>{% t downloads.sourcecode %}</th>
                         </tr>
                         <tr>
-                            <td><img class="small" src="/img/cakewallet.png" width="172" height="202" loading="lazy" alt="Cake Wallet Logo"><a href="https://cakewallet.com/">Cake Wallet</a></td>
+                            <td><img class="small" src="/img/cakewallet.png" width="172" height="202" loading="lazy" alt="Cake Wallet Logo"><a href="https://cakewallet.com/">Cake Wallet *</a></td>
                             <td><span class="icon-android"></span><span class="icon-apple"></span></td>
                             <td>X</td>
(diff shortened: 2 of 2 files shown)


SamsungGalaxyPlayer:
While I understand the sentiment behind this, I don't think it accomplishes the goal.

The issue is that IP address connections are unhidden by default. I think it is arguably worse that some apps use completely random remote nodes without protecting one's IP, which is clearly worse for privacy than just pinging a price API.

Take Monerujo, where one can use their own remote node, but it's not enabled "by default." Feather uses clearnet connections to other nodes for wallet syncing even, ["by default."](https://docs.featherwallet.org/guides/synchronization) I argue Cake's default nodes are just as "trusted" as Feather's default nodes.

If you want to focus on the defaults, one option is changing the asterisk to "Makes clearnet connections to other servers by default, which may be disabled in settings depending on the app." This would realistically cover all the named apps. Every named app here syncs with a remote node over clearnet by default, leaking one's IP.

Practically, what you're probably looking for is "This wallet makes clearnet connections which can't be disabled within the app directly. We recommend using the wallet with system-wide Tor or i2p for the most sensitive uses."

Arguably, then your current applied asterisks are assigned appropriately until Cake Wallet allows disabling of those in a first-time startup process. Thoughts?

erciccione:
@SamsungGalaxyPlayer see https://github.com/monero-project/monero-site/issues/2011#issuecomment-1230000852

nahuhh:
Agree with @SamsungGalaxyPlayer 
There pinging remote nodes is no different.

The issue with the price api (at the time of writing) is it cannot be disabled and ships enabled by default. Meaning, even im using a local node it is still making an external connection.

Mymonero allows LWS afaict, so it *should* be able to be used privately. In practice, I cannot advise as to what connections it makes.

Edge and apps using MyMonero backend are in a whole different of category of "centralized backend".

Open source
1. Best: Private by default. Opt in to connect to remote nodes or servers. (CLI, GUI)
2. Great: tradeoff some privacy for UX. auto connect to remote servers. (cake, monerujo, stack, feather?)
3. Convenience: Forced remote backend (edge)

This being said, _every_ wallet pre v0.18 was making calls to moneropulse

PR #2020, opened 2022-08-15
workflows: check hashes.txt.sig for auto updater

selsta:


binaryFate:
In Install dependencies, the version is hardcoded

.github/workflows/auto-updater.yaml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  validate-auto-updater:
+    name: Validate Auto Updater
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v1
+      - name: Install dependencies
+        run: |
+          sudo apt -y install xvfb libxcb-icccm4 libxcb-image0 libxcb-keysyms1 libxcb-randr0 libxcb-render-util0 libxcb-xkb1 libxcb-shape0 libxkbcommon-x11-0
+          wget -nv https://dlsrc.getmonero.org/gui/monero-gui-linux-x64-v0.18.0.0.tar.bz2
```
binaryFate, line 19: $version

selsta: We need an older version so that we can check if the update is compatible with that older version. It doesn't matter if it is hardcoded, it basically checks can v0.18.0.0 users verify the update.

The current version gets downloaded on line 26.

plowsof (approved):

.github/workflows/auto-updater.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - 'downloads/hashes.txt.sig'
+  pull_request:
+    paths:
+      - 'downloads/hashes.txt.sig'
+jobs:
+  validate-auto-updater:
+    name: Validate Auto Updater
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v1
```
erciccione, line 15: v3 is available. Should we switch to that one?

erciccione (approved):

.github/workflows/auto-updater.yaml
selsta: I intentionally use v1, there is no real advantage for us with v2 or v3

PR #2024, opened 2022-08-15
downloads: update min trezor firmware

selsta:
Don't know about Ledger min firmware.


erciccione (approved):
Rubberstamp utACK.

selsta:
#2034

PR #2025, opened 2022-08-15
downloads: add nano s plus firmware

selsta:
I have no idea what I'm doing and why these are even translatable strings in the first place...


erciccione (changes requested):
Changes should be done only for the English file (`en.yml`).

I don't remember why i had decided to add those as translatable strings (maybe for consistency), but i would be ok with taking them out of the language files.

selsta:
@erciccione do I have to take them out of the english one only? or all?

erciccione:
Sorry for the confusing wording. They should be only removed from the English file. As a general rule, we shouldn't touch non-english language files.

selsta:
#2034

PR #2040, opened 2022-09-06
fixed 404 binance links Exchange.

GraoMelo:
fixed url binance exchange.


plowsof:
Note: your web browser may begin auto directing you to the incorrect localised version of binance.com after visiting e.g. the "/en" version (i kept being directed to the BR url) but appears to resolve itself eventually

plowsof (changes requested):

erciccione:
This file is not used and it's about to be removed.

PR #2051, opened 2022-09-24
user guides: remove link and reference to kovri in multisig-messaging-system.md

erciccione:
As reported by throwawayaccount12 on Matrix:

> When you go to the page https://web.getmonero.org/resources/user-guides/multisig-messaging-system.html#the-architecture-of-the-mms and click the Kovri hyperlink, it redirects to a strange site that attempts to force you to install an add-on in your browser. I am using Firefox. If you wait long enough, it will redirect you to a Russian escort page.

I initially wanted to remove only the link, but since kovri is long time gone, makes sense to remove the entire paragraph.

Note that this document probably needs a complete review, as some parts might be obsolete.


plowsof (approved):
i confirmed that no other occurrences of "kovri.io" exist ACK

rbrunner7 (approved):
Thanks for the PR with this correction, you were faster than me :)

PR #2055, opened 2022-10-10
correct global variable "outdated" text

jorgesumle:
I didn't understand the sentence while translating, so I thought there was a mistake. Is this what you mean?

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -13,7 +13,7 @@ global:
   terms: Terms
   privacy: Privacy
   untranslated: This page is not yet translated. If you would like to help translate it, please see the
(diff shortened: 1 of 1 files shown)


plowsof:
Thanks good spot. This commit should only be "the" -> "it" and contains unnecessary diffs because of white space editing 

we could even delete "the" all together and have "this page is outdated and might not be valid anymore."

jorgesumle:
I think the trailing whitespaces are problematic. See this string from Weblate, it has two spaces between words instead of one:

double-space

`configuration and  tips`
`network  in`

erciccione (changes requested):
This seems to be a duplicate of #2057

PR #2061, opened 2022-10-12
devguides: add prune_blockchain to daemon rpc

plowsof:
pruning_seed? block when pruning began? unsure
check true/false? need to investigate as i didn't notice any difference in the response.


resources/developer-guides/daemon-rpc.md
```diff
@@ -1427,6 +1428,39 @@ $ curl http://127.0.0.1:18081/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"g
 ```
 
 
+### **prune_blockchain** 
+
+
```
selsta, line 1433: Same here

selsta (approved):

selsta (approved):

PR #2082, opened 2022-10-19
Fix links and names in user-guide post

nappex:
Some of links in user-guide post https://www.getmonero.org/resources/user-guides/verification-allos-advanced.html were invalid and doesn't work correctly.

Then I changed also the name of link because the number name which is nowhere in article is confusing for readers in my opinion...


erciccione (changes requested):
Please only edit the english file

erciccione (changes requested):
Please squash your commits into only 1, then we'll be ready to merge this

nappex:
Done. May I ask you why you want to fix only english version?

erciccione (approved):
@nappex translations are dealt with by a script.

PR #2083, opened 2022-10-19
Update docs

plowsof:
closes #2022 
closes #2026 
closes #1842 
closes #1528 
closes #2075 
closes #2084 
closes #2021 
closes #2089 
closes #2093 
closes #2098 

Note: set_log_level / categories are not documented in wallet rpc - but they are in daemon rpc -i think it would be best to list the commands in wallet rpc index but simply link people to the description [like this](https://www.getmonero.org/resources/developer-guides/daemon-rpc.html#set_log_level) , then if something is updated, it only has to be set in one place. 
add_aux_pow is a wip but should not block this PR

PR #2097, opened 2022-11-03
'openalias_address' is a string, not a bool in doc

ryn0:
When getting a response back for validating an address with the RPC client, the result for 'openalias_address' is a string, not a boolean.


plowsof:
confirmed. Thanks, i will add this to #2083 and credit you (the order of the outputs is not important)

ryn0:
Thanks! I didn't think the order mattered but wanted to have the exact response I was getting in the doc.

plowsof:
could we say something slightly different like?
* *openalias_address* - string; Address which the [OpenAlias-formatted address](https://getmonero.org/resources/moneropedia/openalias.html) points to, if given.

plowsof:
closed by #2083 (added as a commit there) thanks ryn0, this can be closed now.

PR #2104, opened 2022-11-22
localization: po/pot files for remote_node_gui guide

plowsof:
Removed whitespace at the end of one in ctrl+f: " \`wallet mode\`. " (appeared to be messing with some string replace) - allowed us to pull in (new) ru translations and not break the norwegian ones.

set 'is translated' to yes on the Russian page.
po4a version 0.68

### DIFF
--- a/_i18n/ar/resources/user-guides/weblate/remote_node_gui.po
+++ b/_i18n/ar/resources/user-guides/weblate/remote_node_gui.po
@@ -7,7 +7,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2021-03-28 11:28+0200\n"
+"POT-Creation-Date: 2022-11-20 16:18+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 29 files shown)


_i18n/en/resources/user-guides/remote_node_gui.md
```diff
@@ -26,7 +26,7 @@ The main menu (`Welcome to Monero` screen) will open. At the bottom left, click
 
 ## Finding a public remote node
 
-First, you will need to find a public remote node to connect to. The website [moneroworld.com](https://moneroworld.com/#nodes) has some great resources for finding nodes. One of the easiest methods would be to use a public remote node run by moneroworld, but they have a tool for finding random nodes too.
+First, you will need to find a public remote node to connect to. The website [moneroworld.com](https://moneroworld.com/#nodes) has some great resources about remote nodes, and the website [monero.fail](https://monero.fail) has an actively maintained list of functioning remote nodes.
```
selsta, line 29: not sure if it matters but there's a trailing space here

plowsof: definitely. this would explain my frustrations with trying to get the translations to match string / replace, thanks! will re-do things after changing

erciccione (changes requested):
I see this PRs edits the language files for #2101 (please make that clear in the future), but that pr has not been reviewed yet and the strings might change. Please convert this to a draft and wait until 2101 is at least approved. So to avoid confusion and unnecessary double works (for both you and reviewers). It's not necessary to rush these changes, better wait a bit but avoid confusion, we can always deploy when we want.

plowsof:
I have dropped the un-reviewed commit (which did not contain the changes from @erciccione) - this was an error on my part, apologies. the localization commit remains unchanged. i have then string replaced the "an actively maintained" -> "a" change.

erciccione (commented):
Please squash

plowsof:
ACK - will squash / resolve conflicts tomorrow

_i18n/ar/resources/user-guides/weblate/remote_node_gui.po
```diff
@@ -17,158 +17,129 @@ msgstr ""
 "Content-Transfer-Encoding: 8bit\n"
 
 #. type: Plain text
-#: _i18n/en/resources/user-guides/remote_node_gui.md:2
+#: ../_i18n/en/resources/user-guides/remote_node_gui.md:2
```
erciccione, line 20: Are these changes of path expected?

plowsof: i believe so, it is what happens when the po-update.sh script is used. 
 the "../" is in relation to the po-update.sh file https://github.com/monero-project/monero-site/blob/master/po/po-update.sh 

an example of this here:
https://github.com/monero-project/monero-site/blob/5c15000bc3d3980a13d2d1a73d8c2364d19967cc/_i18n/ar/resources/moneropedia/weblate/account.po#L26

erciccione (approved):

PR #2112, opened 2022-12-07
roadmap: add seraphis/jamtis wiki link

plowsof:
+ Jamtis / Seraphis wiki link

note: i've noticed a little error of "checked" being added twice.. can edit that for the 2023 bump

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -806,7 +806,7 @@ roadmap:
   haveno: Haveno
 
   ringsize-sera-tryp: increase ring size to >100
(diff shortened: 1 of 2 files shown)


erciccione (commented):
Would be good to add something to the 2023 tab before merging this. You also set it as the default tab for visitors, so there should definitely be something in there or people will be taken to an empty page.

I'd say we can add at least "Haveno" and "Monero <-> Ethereum Atomic Swaps" to the 2023 tab.

erciccione (approved):

PR #2114, opened 2022-12-19
research-lab: Fix link to original whitepaper

davemarchevsky:
I noticed that the link to the original whitepaper was broken - https://cryptonote.org/ appears to be down.

Wikipedia's CryptoNote page uses an Internet Archive link to reference the original paper, while Monero's links to https://bytecoin.org/old/whitepaper.pdf . This patch changes link to Internet Archive, but happy to revise to use bytecoin.org

### DIFF
--- a/resources/research-lab/index.md
+++ b/resources/research-lab/index.md
@@ -180,7 +180,7 @@ permalink: /resources/research-lab/index.html
                         <div class="col">
                             <h2>{% t research-lab.cryptonote %}</h2>
                             <div class="whitepaper">
-                                <a href="https://cryptonote.org/whitepaper.pdf">{% t research-lab.cryptonote-whitepaper %}</a>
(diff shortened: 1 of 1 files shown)


selsta (approved):

erciccione (commented):
The archived page should be reliable enough, but might worth to just host the PDF on getmonero and avoid external links. Thoughts?

plowsof:
> The archived page should be reliable enough, but might worth to just host the PDF on getmonero and avoid external links. Thoughts?

To avoid this problem in the future? sure (getmonero already hosts some pdf's, why not)

erciccione:
Thanks for pointing out the issue @davemarchevsky. I opened #2120, which adds the pdf locally and links to it.

erciccione:
Superseded by #2120

PR #2117, opened 2022-12-24
added monerica and xmr.directory to merchant directory list

HardenedSteel:
Added directories are:
* https://xmr.directory/
* https://monerica.com/

### DIFF
--- a/community/merchants/index.md
+++ b/community/merchants/index.md
@@ -17,6 +17,7 @@ permalink: /community/merchants/index.html
         <ul class="logo">
           <li><a href="https://cryptwerk.com/pay-with/xmr/">cryptwerk.com</a></li>
           <li><a href="https://www.acceptedhere.io/catalog/currency/xmr/">acceptedhere.io</a></li>
+          <li><a href="https://monerica.com/">monerica.com</a></li>
         </ul>
       </div>
     </div>


erciccione (commented):
We could remove `xmr.directory` from this pr in the meantime and merge it with only `monerica`

plowsof:
> We could remove `xmr.directory` from this pr in the meantime and merge it with only `monerica`

agree

erciccione:
ping @HardenedSteel

erciccione (approved):

plowsof:
For visibility: a complaint was made regarding the "open sourceness" of monericas index.html.file (deployed version differing from the one on GH - more recent merges appear but older texts in the src not)

MajesticBank:
I am currently against this proposal to be merged since the website shadow removed me and removed ability to see in transparent way which site is listed and de-listed, why and how.

They basically went closed source.

https://github.com/monerica-project/monerica/issues/129

Claming site is offline:

https://www.uptrends.com/tools/uptime?toolRequestGuid=7d880f58-1ff5-4cfc-b0bc-084fe0d184d5
https://www.host-tracker.com/en/ic/3/e03a4ad9-bf71-4080-aafe-faac8be7b615

However keeping https://www.morphtoken.com/ which is out of service for several years and https://sideshift.ai/ who even doesn't support Monero anymore.

erciccione (commented):
@HardenedSteel please change the tiltle of the pull request to reflect the content of the commit.

My approval was only technical and probably adding this to the merge queue was premature. There seems to be delays in delisting services that don't offer Monero anymore (even for years) and the case of Majesticbank being delisted because the website was temporarily down gives the feeling of listing and delisting based on personal preferences of the individual running the service.

The only bar for quality we are aware of is this claim from @plowsof : "i tried to add a merchant to their page and the maintainer required proof", which is  a personal feedback and not a strong guarantee of quality.

On another note, i don't like the choice of removing the home page from the repository and replace it with a system of multiple files and an audit system that is much more unconvenient to follow and track.

I'm also not a fan of listing a website who puts such political statements at the top of their page:

> The goal of Monerica is to reflect the freedom ideals of the United States of America

For all the reasons above, I'm against adding this website to getmonero.

More feedback is needed.

erciccione:
@johnr365 The difference between monerodocs, a platform meant to integrate technical Monerodocumentation, and monerica, an aggregator of for-profit (but not only) merchants, is fundamental. One is about technical or user documentation, that in the worst case will just not work, the other might create huge problems to a final user.

monerodocs is documentation, monerica is a collection of many links to third party services/software. The two services and the risks related in case of suboptimal maintenance are completely different and not really comparable. Especially if considered that one is in a dev-specific section of getmonero and the other one is in the arguably most sensitive page of the website (the merchants page).

> it's appearing down for me. Is it finished? And if so, is there another PR to remove it?

Looks down to me as well. If we are sure it's not going up again, we should remove it.

HardenedSteel:
@erciccione:
> On another note, i don't like the choice of removing the home page from the repository and replace it with a system of multiple files and an audit system that is much more unconvenient to follow and track.
> 
> I'm also not a fan of listing a website who puts such political statements at the top of their page:

I agree.

If we won't add [Monerica](https://monerica.com/) either it's better to close the pull request.

PR #2125, opened 2023-02-02
tags: add missing for seraphis blog

### LINKED ISSUE #2124: bug when displaying tags on blog posts
Screenshot_20230202_100231 
(from https://www.getmonero.org/2023/02/02/seraphis-jamtis-developer-opportunities.html)

There is some problem while looping through the tags of a post.  I might have an idea of why, as it could be related to the trimming of blog post tags done some years ago.

plowsof:
closes #2124

### DIFF
--- a/_data/tags.yml
+++ b/_data/tags.yml
@@ -87,3 +87,6 @@
 
 - slug: urgent
   name: Urgent
+
+- slug: seraphis
+  name: Seraphis


erciccione (commented):
I'm not sure if we should introduce these tags. We removed them all in pasts becasue we had a lot of tags and most of them were used in one single post. This resulted in bloat and the tag becoming basically useless.

That's why we opted for keeping only the tags that had their own menu entry in the blog (but we left some tags of old blog posts). Personally i would keep that approach and would get rid of the extra tags in the seraphis blog post instead. We could decide which ones of those tags might be used in the future and keep those, but i wouldn't introduce new tags that will very likely be related to a single blog post.

plowsof:
I agree, if there are future blogs where these tags should begin appearing and become relevant for 'searching' then sure. 

i think only the Seraphis tag would be required? (a blog with jamtis mentioned would fall under the seraphis umbrella)

erciccione (approved):

HardenedSteel (changes requested):
Don't you need to ~~add other missing tags or~~ remove the missing tags from the blog post? Or do we need to do another PR for that? The issue is currently not resolved. 
image

plowsof:
@HardenedSteel a PR to delete the other tags from the blog post is the next step now that this is approved :+1:

HardenedSteel (approved):

PR #2129, opened 2023-02-18
remove merchants.yml file

HardenedSteel:


plowsof:
is this file currently in use somewhere? is there somewhere with the entire list on site? (removing localmonero will raise some eyebrows)

HardenedSteel:
@plowsof 
> is this file currently in use somewhere? is there somewhere with the entire list on site?

I searched up but couldn't find anything. If not needed we may delete it.
> removing localmonero will raise some eyebrows

As I said, Agoradesk and Localmonero are basically the same site, Agoradesk additionally has BTC support. Maybe keeping only Localmonero and not Agoradesk can be better, but I don't see any reason to keep both sites.

erciccione (changes requested):
please squash

erciccione (approved):

PR #2130, opened 2023-02-21
add information about sites' I2P support

HardenedSteel:


erciccione (changes requested):
Would be good to make clear in the commit description that the changes are meant for the merchants page. For example: `merchants: add information about sites' I2P support`

erciccione (approved):

PR #2136, opened 2023-03-03
moneropedia: remove entry canonically unique host

### LINKED ISSUE #1308: Delete or simplify Moneropedia entry 'Canonically unique host'
Is it really useful for the user? If it is, would be better to simplify it.

We can start working on this after #1270 is resolved.

erciccione:
closes #1308


plowsof (approved):

plowsof:
Confirmed that its removed from the moneropedia page, thanks. A follow up PR can remove other instances of canonically (e.g. the weblate po files?) but not important

erciccione:
@plowsof good point. Removed them and pushed.

plowsof (approved):

PR #2142, opened 2023-03-27
add moneropedia to po script

erciccione:
Recovered and rebased #1984

### DIFF
--- a/po/po-update.sh
+++ b/po/po-update.sh
@@ -2,22 +2,22 @@
 
 # SCRIPT
 #
-# This script loops through all the .config files in this folder and run po4a on them. The config files are one for each user guide.
+# This script loops through all the .config files in this folder and run po4a on them. The config files are one for each user guide and moneropedia entry.
 # See the next section for instructions about adding new config files.
 #
 # CONFIG FILES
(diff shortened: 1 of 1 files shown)


plowsof (approved):

po/po-update.sh
```diff
@@ -37,6 +37,10 @@ updatelangs() {
     echo "-> Running po4a on ${guide}"
     po4a $guide || echo "ERROR: something went wrong, po4a didn't run succesfully on ${guide}"
   done
+  for entry in moneropedia/*.config; do
+    echo "-> Running po4a on ${entry}"
+    po4a $entry || echo "ERROR: something went wrong, po4a didn't run succesfully on ${entry}"
+  done
 }
 
 while true; do
```
hinto-janai, line 46: I know this was already here but this `while` does nothing, it immediately `break`'s. It can be removed.

hinto-janai (approved):

po/po-update.sh
plowsof: missed this comment, true, its not looping anything (had to look at the entire file to see)

PR #2158, opened 2023-04-20
hangouts: update matrix links

plowsof:
MRL has a monero.social room

### DIFF
--- a/_data/chats.yml
+++ b/_data/chats.yml
@@ -18,7 +18,7 @@ chats:
   matrix: https://matrix.to/#/#monero-pools:monero.social
 
 - monero-research-lab:
-  matrix: https://matrix.to/#/#monero-research-lab:matrix.org
+  matrix: https://matrix.to/#/#monero-research-lab:monero.social
 
 - monero-translations:
   matrix: https://matrix.to/#/#monero-translations:monero.social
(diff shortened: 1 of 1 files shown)


erciccione (approved):
Note that the change of address is only necessary for `monero-site:haveno.network`, since that server is now offline and the address is unreachable. For this we can either use `matrix.org`, which is the original server or `monero.social`. it doesn't make any practical difference.

The address of the research lab is working. It was registered on matrix.org, but we can use monero.social.

It doesn't make any practical difference (as long as both servers stay online), but since we are changing it i would use the same server address for both rooms. We cannot use `monero-site` on the matrix server, so maybe better use monero.social, so we can use `monero-site:monero.social`

plowsof:
agreed, much better to to be on monero.social. and to avoid -site / -website confusion

erciccione (approved):

hinto-janai (approved):

nahuhh:
There was a suggestion to remove -hardware.
Would that be within the scope of this PR, or should a separate pr's be opened for removals and / or additions?

nahuhh (approved):

plowsof:
if removing the hardware channel is decided, there is a loop in the hangouts html. which first gets all the hangouts: from en_yml
https://github.com/monero-project/monero-site/blob/c3c21f92cc62c7cdba5133cc8854e64a4c97a4bc/_i18n/en.yml#L117
and then from that list ^ it finds the links from the edited file in this PR.  deleting it from en,yml would work, but leave trash in this file so both end up being edited, i assume a separate PR that would use this one as the first commit would be better

PR #2164, opened 2023-05-06
Fix grammatical issue on home page

getsnoopy:


erciccione (changes requested):
Is this a grammatical issue? Seems more like a preference. Doesn't worth changing in my opinion, especially if we consider that translators will be asked to re-translate it.

getsnoopy:
It's not a preference at all, but a grammatical issue. The second phrase is not explaining how or why Monero is a currency, but an independent clause that's related to the previous clause. That calls for a semicolon, not a colon.

erciccione (approved):

PR #2165, opened 2023-05-07
Update Library text to use consistent date format

getsnoopy:


erciccione (commented):
Why is this a consistency issue? I don't see other dates. If there is no inconsistency i don't see the point of the change.

getsnoopy:
There are dates on the Blog page, and regardless, the site should be using the internationally standard RFC 2822 date format.

erciccione (approved):

PR #2169, opened 2023-05-14
faq: change categories

### LINKED ISSUE #2006: Proposal: Sort the FAQ by more specific themes
The FAQ has grown over the years, and with 38 questions, it needs more structure than the three current categories General, Advanced and Node and Wallet. The first two categories are meaningless anyway. After I saw a PR for a script that sorts the questions alphabetically, at least in English, in #1969, I thought there must be a better way. Let's use more specific categories, so we can group a handful of Q&A's in each, while users can quickly find the topic they're interested in.

What do you think of the following structure? I also cleaned up the grammar here and there.

## Basics

* How can I get Monero?
* Why is Monero called 'Monero'?
* How is Monero different from Bitcoin?
* What's the meaning of [technical word]?
* Are there videos I can watch to learn about Monero?
* Is it true that Monero has a hard fork every 6 months?
* Does Monero have a block size limit?
* How can I contribute?

## Privacy

* How is Monero’s privacy different from other coins?
* Is Monero magic and does it protect my privacy no matter what I do?
* Is Monero 100% anonymous?
* Is Monero a mixer or mixing service?
* (Proposed in #1766:) What is Dandelion++?
* (Proposed in #1887:) What if my transaction was mined in a block containing only two transactions?

## Security

* Why is my antivirus/firewall flagging the Monero software I just downloaded as malware?
* What is ASIC resistance? Why is it important?
* Are there known vulnerabilities in Monero?
* (Proposed in #1783:) I've been hit by a ransomware and they want me to pay a ransom in Monero. What can i do?

## Economics

* How does Monero have value?
* What is fungibility, and why is it important?
* What is Monero's maximum supply?
* (Proposed in #1965, maybe redundant:) Why doesn't Monero have a maximum supply?
* If Monero is so private, how do we know coins aren't being created out of thin air?
* (Proposed in #1792:) Does Monero support Atomic Swaps?

## Wallets

* Which wallet should I use?
* What is the difference between a light wallet and a normal wallet?
* I can't see my funds. Did I just lose all my Monero?
* I haven't touched my Monero in a long time, did I lose my coins due to a network upgrade?
* Why is my wallet taking so long to sync?
* Why does my wallet need to scan the blockchain everytime I open it?

## Node

* How do I decide if I should run a full node or a pruned node?
* How big is the Monero blockchain?
* Why does the blockchain need so much space?
* Can I avoid downloading the entire blockchain?
* Can I manually import the blockchain?
* How can I connect my node via Tor?
* Is it dangerous to run a personal node?
* Is it dangerous to use a remote node? What's the data a node operator can get from me?

erciccione:
As suggested by @apertamono in #2006

Closes #2006

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -443,11 +443,14 @@ contributing:
   altqrbitcoin: Bitcoin QR code
 
 faq:
+  basics: Basics
+  privacy: Privacy
+  security: Security
+  economics: Economics
+  wallets: Wallets
+  node: Node
   intro: Find answers to recurring questions and concerns about Monero.
   toc: Table Of Contents
-  general: General
-  advanced: Advanced
-  nodeandwallet: Node and Wallet
   additional: "More Info:"
   resandhelp: Resources & Help
   stackexchange: One of the most complete resources about Monero. If you have a question which is not in this FAQ, you will likely find the answer on the Monero StackExchange.
--- a/get-started/faq/index.md
+++ b/get-started/faq/index.md
@@ -19,46 +19,55 @@ meta_descr: faq.intro
                     </div>
                 </div>
                 <ul>
-                    <li class="category">{% t faq.general %}</li>
+                    <li class="category">{% t faq.basics %}</li>
                     <ul class="logo">
                         <li><a href="#anchor-word">{% t faq.qword %}</a></li>
                         <li><a href="#anchor-contribute">{% t faq.qcontribute %}</a></li>
-                        <li><a href="#anchor-value">{% t faq.q1 %}</a></li>
                         <li><a href="#anchor-buy">{% t faq.q2 %}</a></li>
-                        <li><a href="#anchor-different">{% t faq.q4 %}</a></li>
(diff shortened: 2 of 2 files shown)


plowsof:
Thanks for completing this. Seems a lot better now.  I noticed a small anchor issue (not caused by this PR, maybe out of scope) "Is it true that Monero has a hard fork every 6 months?" links to #monero-meaning instead, and other anchors have "anchor-" in their names / others don't, seems a bit redundant.

plowsof (approved):

HardenedSteel (changes requested):
The FAQ `Is it dangerous to use a remote node? What's the data a node operator can get from me?` is missing.

erciccione:
> The FAQ Is it dangerous to use a remote node? What's the data a node operator can get from me? is missing.

Fixed

apertamono:
@erciccione Thanks for picking this up. I never implemented it because I didn't get around to asking for more feedback in the chat channels.

HardenedSteel (approved):

PR #2178, opened 2023-06-14
Translations from weblate

Monero-Weblate:
Bulk of French translations from multiple contributors, but also translations for Turkish and Chinese


plowsof:
are we removing the @canonically-unique-host place holders from Moneropedia -> Locally-unique host also? (or the entire guide)

_i18n/fr.yml
```diff
@@ -89,7 +90,7 @@ home:
     confidentialité et avec des frais minimes, contre des biens ou des services. Nombre
     d'entités  acceptent le Monero (XMR), regardez notre liste de commerçants qui
     acceptent actuellement le Monero.
-  useit: "C'est une monnaie : utilisez-la !"
+  useit: "C'est une monnaie; Utilisez-la !"
```
plowsof, line 93:
```suggestion
  useit: "C'est une monnaie; Utilisez-la!"
```
maybe not an issue but i noticed the "!" on a new line because of screen width and at least some other translations of no space between the "!"

plowsof:
noticed an @démon under Documentation RPC on the dev guides page  https://deploy-preview-2178--barolo-time-757cf9.netlify.app/fr/resources/developer-guides/index.html

_i18n/fr.yml
erciccione: In French they put a space before question and exclamation marks, but i don't think it's an issue if it's inconsistent here and there. More annoying to have the ! in a newline, but doesn't worth changing structure because of it. Will look what can be done.

erciccione:
> are we removing the @canonically-unique-host place holders from Moneropedia -> Locally-unique host also? (or the entire guide)

We can do it independently from this PR. It's better to not manually touch strings coming from Weblate

plowsof (approved):

PR #2186, opened 2023-07-23
devguides: fixes to the get_address descriptions

dimalinux:
Fixes incorrect descriptions of the encoding used for addresses. Let's the user know that the primary address is also returned. Includes information on the default behavior when the `address_index` field is omitted. (Note that `address_index` should have been named `address_indices`, as it is in the outputs of `create_address`, but that's not something the documentation can fix.)


erciccione (commented):
Thanks @dimalinux. Please squash

erciccione (approved):

PR #2191, opened 2023-08-21
remove double quote to avoid breaking img in press kit

erciccione:
Weblate replaces single quotes in this string with double quotes, breaking the html. I'm removing quotes altogether to avoid the issue.

*note that usually language files shouldn't be edited from Github. I'm doing it here because the images will look broken otherwise*

### DIFF
--- a/_i18n/de.yml
+++ b/_i18n/de.yml
@@ -488,7 +488,7 @@ press-kit:
   whyconf1: Wir empfehlen jedem, Community-Mitglieder aus dem größten Datenschutzprojekt
     in verwandte Nachrichtenartikel und Veranstaltungen einzubeziehen.
   altpk-2: Bild von fluffypony während einer Präsentation vor Publikum auf der DEFCON
-  altpk-1: Bild einer Frau, die vor einem "Monero Talk"-Banner steht
+  altpk-1: Bild einer Frau, die vor einem Monero Talk-Banner steht
   whycommunity1: Monero ist für Transaktionen bekannt, die weitaus privater sind als
     bei allen anderen Kryptowährungen zusammen.
   contactsp: Es gibt mehrere Möglichkeiten, mit der Community in Kontakt zu treten.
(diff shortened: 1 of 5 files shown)


_i18n/en.yml
```diff
@@ -343,7 +343,7 @@ press-kit:
   whycommunity1: Monero is trusted for more private transactions than all other cryptocurrencies combined.
   whyconf: The Monero community has substantial involvement at Defcon, Grayhat and the CCC (Chaos Computer Club). The Monero Konferenco, an academic conference, is hosted annually.
   whyconf1: We encourage everyone to include community members from the largest privacy project in related news articles and events.
-  altpk-1: Picture of a woman standing in front of a 'Monero Talk' banner
+  altpk-1: Picture of a woman standing in front of a Monero Talk banner
```
rbrunner7, line 346: Well, single apostrophes instead of double qoutes do work, but I see how you would like to achieve consistency over all languages.

_i18n/de.yml
```diff
@@ -488,7 +488,7 @@ press-kit:
   whyconf1: Wir empfehlen jedem, Community-Mitglieder aus dem größten Datenschutzprojekt
     in verwandte Nachrichtenartikel und Veranstaltungen einzubeziehen.
   altpk-2: Bild von fluffypony während einer Präsentation vor Publikum auf der DEFCON
-  altpk-1: Bild einer Frau, die vor einem "Monero Talk"-Banner steht
+  altpk-1: Bild einer Frau, die vor einem Monero Talk-Banner steht
```
rbrunner7, line 491: Just a small nitpick, but in German it would look better with two hyphens like so: "Monero-Talk-Banner"

rbrunner7 (approved):
Never mind my two comments about small issues, the PR looks good to me.

_i18n/en.yml
erciccione: It's not really about consistency. Weblate replaces single quotes with double quotes and people might not pay attention and use double quotes when they shouldn't, resulting in a broken page. Removing the quotes will avoid the problem.

_i18n/de.yml
erciccione: This format was chosen by german translators, we can always suggest a new translation on weblate (after it will be up again) :)

rbrunner7: Ok. Has nothing to to with the problem the PR wants to solve anyway!

PR #2194, opened 2023-10-07
downloads: add GUI macOS ARM link

selsta:


plowsof:
small edit of the -lt value on line 60 needed "elif [ $num -gt 13 ] && [ $num -lt 19 ] ; -> -lt 20

i am not 100% happy with the filenames checker - it will error - however the error message can be misleading - but most importantly - it will error if there is a problem.  this LGTM

plowsof:
Validate hashes error will be fixed after @binaryFate adds the new gui arm8 hash/filename to hashes.txt

plowsof (approved):

PR #2197, opened 2023-10-11
workflows: verify redirect urls in downloads.yml

plowsof:
re-download all binaries using their redirect urls and confirm hashes.

displays a warning if files downloaded using the redirect urls do not match the hashes in downloads.yml

if there is a mis-match / or the redirects are not updated to point at the new files yet, the workflow will still be green and only produce warnings.  re-running the workflow manually after site is deployed / redirects pointing to new files will remove the warnings.

example run showing a warning: https://github.com/plowsof/monero-site/actions/runs/6487887488

on `workflow_dispatch:` can be added to make running the workflow manually (if needed)

### DIFF
--- a/.github/workflows/hashes.yaml
+++ b/.github/workflows/hashes.yaml
@@ -119,6 +119,7 @@ jobs:
           grep monero- downloads/hashes.txt | sha256sum -c
       - name: Verify downloads.yml hashes
         run: |
+          mkdir verify_redirects
           yq -r '.[] | .[0].downloads[] | "\(.link)|\(.hash)"' _data/downloads.yml | grep -v github |
               while read line; do
                   [ -z "$line" ] && continue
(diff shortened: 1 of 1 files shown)


plowsof:
The problem has been fixed. This workflow will prevent/detect the problem we had, however, 
- it needs to be run only when the DNS is updated (which comes some time after deployment) so we have to run the whole thing twice.
- alot of duplication if we have a separate workflow for this check.
- re-downloading all the files can add extra chances of network timeouts / the script failing (which happens often already)

i think it's better suited to be run on a contributors repo , 'after the fact' to confirm things are ok, rather than here on -site.. im leaning to close

erciccione:
> i think it's better suited to be run on a contributors repo , 'after the fact' to confirm things are ok, rather than here on -site.. im leaning to close

Agree. Or a script on somebody's computer. No reason to add it on this repo.

PR #2198, opened 2023-10-14
Rm paragraph about locked outputs being used as decoys

jeffro256:
Locked outputs cannot be used as ring members since the protocol (by design) has no way of knowing which output is the "true" spend inside of a ring signature. Thus, the protocol only validates ring signatures where each member of the ring is properly unlocked.


erciccione (changes requested):
Only the original file in english should be edited (`_i18n/en/resources/moneropedia/unlocktime.md`)

jeffro256:
Should I modify the commit now?

erciccione (approved):
Would be better to improve the commit description to specify what is being changed, but it's good to go.

jeffro256:
Is this commit message better?

erciccione (approved):
That's perfect, thanks.

PR #2213, opened 2023-12-15
refactor: add 'Trezor Safe 3' to HWs that support monero

tohrxyz:
Hey, I've noticed that in section about hardware wallets that currently support Monero, from Trezor only Model T is mentioned. But the new model that they released called [Trezor Safe 3](https://trezor.io/trezor-safe-3) also supports Monero as you can read here 

image

Therefore, I've added new item to the list of Trezor wallets that support Monero, hopefully this small improvement will help people who are searching for this information.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -360,6 +360,7 @@ meta_descr: downloads.intro
                                     <td>
                                         <ul>
                                             <li>Model T</li>
+                                            <li>Safe 3</li>
                                         </ul>
                                     </td>
                                     <td>


hinto-janai:
I believe this has to wait until https://github.com/monero-project/monero-gui/pull/4255 is merged.

The line that appears above has to be changed as well:

https://github.com/monero-project/monero-site/blob/cd8ed0ebc31a96d3de3902fa734a2dc6f844d542/_i18n/en.yml#L297

PR #2218, opened 2024-01-11
Fix broken link

jorgesumle:
You could put a link like `https://www.binance.com/en/crypto/buy/EUR/XMR`, but I don't like this approach, because the URL might break again in the future, it forces the English language in the URL, and you have to pick a currency.


nahuhh:
would like this closed in favor of #2223
@jorgesumle please see #2223

jorgesumle:
I agree! Fuck Binance

PR #2237, opened 2024-02-01
ruby: bump version

plowsof:
dropping support for ruby < 3.0.0
ruby version 2.7.8 is EOL: https://www.ruby-lang.org/en/downloads/

v3.1.4 is the latest stable version that will work with netflify 

ci: added v3.0.2 as this is a common ruby version (e.g installed by default on my ubuntu distro)


nahuhh:
Debian 12 is using 3.1, so ubuntu LTS 24.04 likely will be too

plowsof:
changes:
- netflify previews will mirror the ruby version on the getmonero server 3.0.2 
- 3.1.0 (default in [debian 12 bookworm](https://packages.debian.org/bookworm/ruby) now used in gh workflows

plowsof:
3.0 was EOL'd 2 months ago https://endoflife.date/ruby 

3.1's support ends in 8 months 

> Since Ruby 2.1, a new major version of Ruby has been released every year on December 25th, and EOLed 3 years, 3 months later.

plowsof:
- moving jekyll-multiple-languages plugin (v1.8.0) to local so we can modify it
- fix the small issues causing errors on ruby versions >3.2.0
- we should update getmonero to the latest stable 3.3.4 (2~ years of support)

the 1st commit adding the local plugin should have the same sha256 as what exist on [master](https://github.com/kurtsson/jekyll-multiple-languages-plugin/blob/master/lib/jekyll-multiple-languages-plugin.rb)

we can kick the can a little further down the road now and remain on jekyll.

nahuhh (approved):

PR #2241, opened 2024-02-03
downloads/faq: remove blockchain.raw

plowsof:
current version on getmonero as of making this issue: (its has been updated since)

>71697681675 Feb 23  2023 blockchain.new

benefits of this PR:
- reduction in resources for getmoneros server (bandwidth/disk usage)
- make people verify their own blockchain / decentralisation / reduced liability for getmonero , or any other buzzwords that may support this  
- often times users run into problems when beginning with an imported blockchain or it simply prolonged their suffering before seeking help

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -226,12 +226,7 @@ downloads:
   choose: Choose your download
   gui: Monero GUI Wallet
   cli: Monero CLI Wallet
-  blockchain: Import Blockchain
-  blockchain1: If you'd prefer to use a raw blockchain instead of syncing from scratch, you can use the most current bootstrap. It is typically much faster to sync from scratch, however, and it also takes a lot less RAM.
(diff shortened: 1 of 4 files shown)


HardenedSteel:
Possible use case for importing blockchain.
https://monero.stackexchange.com/a/14286/12826

plowsof:
Use case: "run a node without validating whole Monero blockchain? This can be useful for who wants to run a node on slower disks like HDDs."

description of flag `--dangerous-unverified-import` [here](https://github.com/monero-project/monero/pull/3948/files)
>Blindly trust the import file

Nodes who have not validated the blockchain, who download from a central source, are not good for the network, so i don't see this as a use-case we should support. If a large majority of the network where weak HDD machines that have been essentially propped up by importing a non-verified blockchain then during times of huge transaction volume / spam they would crumble.

selsta (approved):

nahuhh (approved):

PR #2244, opened 2024-02-08
faq: add tail emission/nodes video

plowsof:


nahuhh (commented):
`Monero_Tail_Emission-VOSTO_EMISIO.mp4`


please make a habit of using - and _ and not [space] for filenames. 🙏😭

nobody likes %20 in their filename

plowsof:
im not willingly putting spaces / %20 in filenames (and having to forcepush about 5 times). when a pull request is made, you generally try to blend in / use similar code styles/syntax as the area you are changing if there are no strict standards in place

nahuhh:
Yes yes, but the filename is bad (spelling and an extra space at the end), even if it use %20 like other stuff in the list

plowsof:
spelling/extra space corrected thanks, i requested an m4a and got an mp4, lets wait for vosto emisio to send an m4a (also no idea about these formats)

plowsof:
took a look at what M4V is and.. its not good:
>M4V is a container video format developed by Apple, closely resembling the MP4 format. The main difference between M4V and MP4 is that an M4V file can enjoy DRM copy protection. This DRM protection is in place to prevent piracy, ensuring that M4V files purchased from the iTunes store cannot play on non-Apple players.

we will definitely not be getting an M4V version of the tail emission video from vosto :D

plowsof:
after realising what m4v format is, i dont think we should blend in anymore and stop using m4v. i have also replaced spaces with "_" in the filename. The only concern now could be that the filesize is 40~MB

HardenedSteel:
WebM can be even better.

plowsof:
Thanks Vosto. I've added the 2 webm files. instead of adding a new translated string for nodes, ive recycled one for "Node" hopefully this is OK

plowsof:
fixed a typo `Emsision` -> `Emission`

PR #2251, opened 2024-02-26
Update get-started/mining/index.md link

twizco:
Link update


plowsof (approved):

plowsof:
i notice the CSminer repo is several years old, i wonder if we should simply remove it. and replace it with GUPAX.io

HardenedSteel (commented):
The miner isn't maintained like last 3 years. It may be better to remove.

jermanuts:
> replace it with GUPAX.io

Agree.

PR #2262, opened 2024-03-09
update CCS incident blog post

HardenedSteel:
added information about the Monero General Fund wallet received 2696.73 XMR and strikethroughed the old information


jermanuts:
Add https://monero.observer/2696-xmr-anon-donation-monero-general-fund/

PR #2264, opened 2024-03-09
downloads: add Linux RISC-V link

selsta:
In preperation for v0.18.3.2 adding support for RISC-V.


selsta:
Sorry for multiple force pushes, should be good now. Also why does netlify keep making a new comment instead of editing the original one like previously?

plowsof (approved):

PR #2293, opened 2024-05-08
Remove LocalMonero from list of merchants

cottnn:
LocalMonero will close on November 7th, 2024: https://localmonero.co/blog/announcements/winding-down

### DIFF
--- a/_i18n/de.yml
+++ b/_i18n/de.yml
@@ -179,7 +179,6 @@ merchants:
   locmondescr: Online-P2P-Börse für den Handel von Privat zu Privat.
   hodlhodldescr: P2P-Handelsplattform mit unpfändbarer Treuhand.
   visitbisq: Besuche Bisq
-  visitlocalmonero: Besuche LocalMonero
   visithodlhodl: Besuche HodlHodl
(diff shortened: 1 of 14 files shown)


plowsof:
Thank you. we should only remove the translation from the en.yml file. the others dont need to be modified

nahuhh:
Should likely go into next deploy

new ads = immediately
new accts = immediately
new trades = may 14

the rest is time for disputes, and withdrawing funds

plowsof:
We should.consider leaving it up and displaying the above timeline for 6 months with the deletion coming after as this may help someone

HardenedSteel (approved):

nahuhh:
Are the removed tl strings a problem?

@plowsof

plowsof:
when deleting keys like this, we should _only_ make the change in the english file. weblate then deletes the unused keys (note master branch and subsequent "please squash" request so i can just resubmit this PR if its approved)

plowsof:
replacing with #2321 which adds @HardenedSteels correct deletion method + removes localmoneros block explorer from tools.

PR #2296, opened 2024-05-09
merchants: remove fixedfloat

plowsof:
re-considered for listing after their monero swaps are operational again. currently under maintenance.


HardenedSteel (approved):

jermanuts:
I wouldn't reconsider it tbh they got [hacked twice](https://www.web3isgoinggreat.com/?id=fixedfloat-hack-2) and changed their domain to https://ff.io/ "As a fresh start" or something.

PR #2199, opened 2023-10-16
update internationalization of user guides and moneropedia

erciccione:
- run po4a script on user guides and moneropedia entries
- fix issues in _i18n/es/resources/user-guides/remote_node_gui.md
- removed leftovers of removed guide canonically_unique_host

PR #2300, opened 2024-05-17
Resolve 2199

plowsof:
closes #2199

2199 has merge conflicts with a timestamp in unlocktime.po / .pot files which is fixed.
when cherry picking the commit in 2199 there after, the ring-size in russian translation has to be corrected back to 16.


_i18n/de/resources/moneropedia/pruning.md
```diff
@@ -16,18 +16,16 @@ Pruned-Nodes keine Nachteile bezüglich der Privatsphäre oder Sicherheit.
 
 Es gibt drei Möglichkeiten des Prunings:
 
-- *Synchronisieren eines Pruned-Nodes von Grund auf:* Du synchronisierst und
```
plowsof, line 19: example of "fuzzy" not cleaned up by the original commit author/dropping translations. there are other cases. todo

PR #2306, opened 2024-06-16
add note about `get_transfers`/`get_transfer_by_txid` `destinations` being undefined, make it more consistent

recanman:


resources/developer-guides/wallet-rpc.md
```diff
@@ -1892,6 +1892,7 @@ $ curl http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"c
 Returns a list of transfers.
 
 <p style="color:red;"><b>WARNING</b> Verify that the transfer has a sane <code>unlock_time</code> otherwise the funds might be inaccessible.</p>
+<p style="color:red;"><b>WARNING</b> <code>destinations</code>, only available for outgoing transactions, will be <b>UNDEFINED</b> if not in the wallet cache.</p>
```
jeffro256, line 1895: I think the warning should only be for the fact that you lose this information if this wallet cache was not the one to construct the transaction, since that's the important part. The detail about it only being for outgoing transactions is just part of the interface. 


```suggestion
<p style="color:red;"><b>WARNING</b> The <code>destinations</code> fields are only available when this wallet cache was the one to construct the transaction. If you restore your wallet from scratch, you will lose this information. </p>
```

jeffro256 (approved):

HardenedSteel:
squash?

recanman:
@HardenedSteel squashed

plowsof:
would it not be better to just add a "Note: ..." next to the destimations param? the scary warning for unlock time was an exception, unless this is of the same severity if overlooked?

HardenedSteel (approved):

resources/developer-guides/wallet-rpc.md
```diff
@@ -1984,6 +1985,7 @@ $ curl http://127.0.0.1:18082/json_rpc -d '{"jsonrpc":"2.0","id":"0","method":"g
 Show information about a transfer to/from this address.
 
 <p style="color:red;"><b>WARNING</b> Verify that the transfer has a sane <code>unlock_time</code> otherwise the funds might be inaccessible.</p>
+<p style="color:red;"><b>WARNING</b> <code>destinations</code>, only available for outgoing transactions, will be <b>UNDEFINED</b> if not in the wallet cache.</p>
```
jeffro256, line 1988: Kind of a nitpick, but it would be nice if this message matched the one above.

jeffro256:
> would it not be better to just add a "Note: ..." next to the destinations* param? the scary warning for unlock time was an exception, unless this is of the same severity if overlooked?

I think it's worthy of a warning, since you can permanently lose accounting information, which I remember caused at least one person financial damage since they couldn't provide a valid tx spend proof to their exchange.

plowsof:
with the new docs site we can play around with warnings. 

> > would it not be better to just add a "Note: ..." next to the destinations* param? the scary warning for unlock time was an exception, unless this is of the same severity if overlooked?
> 
> I think it's worthy of a warning, since you can permanently lose accounting information, which I remember caused at least one person financial damage since they couldn't provide a valid tx spend proof to their exchange.

the possible repercussions of deleting the wallet cache / re-syncing a wallet would fill the entire doc with red warnings. so i would NACK this PR as is. This would seem better fitting as a Note about working with wallets / monero-wallet-rpc. The new docs site can facilitate this better i feel. 

- example of notes at the top https://2.test.monerodevs.org/interacting/monero-wallet-rpc-reference/
- different style of a note e.g. warning example @ https://2.test.monerodevs.org/interacting/monero-wallet-rpc-reference/#describe_transfer
- they can be collapsible also

plowsof:
docs are moving to https://github.com/monero-project/monero-docs

merged in https://github.com/monero-project/monero-docs/pull/67

closing. thank you

PR #2309, opened 2024-06-20
hangouts: add warning for matrix.org homeserver

HardenedSteel:
add a warning about matrix.org homeserver is blocking monero.social


HardenedSteel:
after long time later the issue seems to be fixed

PR #2316, opened 2024-07-08
library: remove The Monero Standard newsletter

### LINKED ISSUE #2315: Remove The Monero Standard newsletter
Library>Newsletters

https://localmonero.co/blog/announcements/winding-down

rottenwheel:
Closes #2315.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -950,9 +950,6 @@ library:
   newsletters: Newsletters
   revuomonero: >
     Revuo Monero is a weekly newsletter where you can find the most recent Monero news.
-  themonerostandard: >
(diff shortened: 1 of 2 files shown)


HardenedSteel:
They wrote a lot stuff and completely removing isn't ok to me tbh. Perhaps we should create an archive section?

nahuhh:
> They wrote a lot stuff and completely removing isn't ok to me tbh. Perhaps we should create an archive section?

Localmonero said the info would still be available somewhere. Should clarify where and replace with that link

HardenedSteel (approved):
approving because it can be re-added altogether with a PR about archived newsletters or projects

Issue #2325, opened 2024-07-19
Request: fix roadmap in mobile portrait view 


plowsof:
Viewing the roadmap on mobile in portrait. I vaguely know CSS can grab screen orientation so improvements here should be possible.

- make the current year appear first as done in landscape.
- ++ would be improvements to how the data is shown in this orientation.

Also the year being displayed in the tab and again above the entries. IMO this is only needed for the archived tab with multiple.years and could.be removed.

https://www.getmonero.org/resources/roadmap/

Issue #2324, opened 2024-07-17
css/custom.css file has errors/warnings 


plowsof:
[w3.org analysis](https://jigsaw.w3.org/css-validator/validator?uri=getmonero.org%2Fcss%2Fcustom.css&profile=css3svg&usermedium=all&warning=1&vextwarning=&lang=en)

luish-uk:
Hello, I have read this issue and understand what needs fixing regarding the W3C errors and warnings. I have run the validator myself and I would like to tackle the handful of errors and couple hundred warnings first as a pull request. @plowsof can you please let me know whether you would like one pull request covering the grouped fixes or rather split into smaller ones, also let me know if there's anything already in progress.

PR #2327, opened 2024-07-20
resources: roadmap > upd8 portrait layout

nahuhh:
Closes https://github.com/monero-project/monero-site/issues/2325

- reduced the trigger width for moving from header view to inline view (use header view unless very narrow viewport)
- slightly reduced header font size
- abbreviated the years

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -3639,10 +3639,10 @@ img.emoji {
 
 .tabPanel-widget > label,
 .tabPanel-widget > h2 {
-  font-size: 1.1rem;
+  font-size: 1rem;
   width: calc(100% / 7);
-  height: 2rem;
-  line-height: 2rem;
+  height: 1rem;
+  line-height: 1rem;
   padding: 1rem 0;
 }
 
@@ -3805,7 +3805,7 @@ img.emoji {
 
 }
 
-@media only screen and (max-width: 48rem) {
+@media only screen and (max-width: 24rem) {
     .roadmap .mobile-roadmap {
     height: 2rem;
 }
@@ -3894,7 +3894,7 @@ img.emoji {
 }
 
 .tabPanel-widget h2 {
-    font-size: 1.4rem;
+    font-size: 1rem;
 }
 
 .roadmap .tabPanel-widget > input + h2 + div.tabPanel-content {
--- a/resources/roadmap/index.md
+++ b/resources/roadmap/index.md
@@ -28,7 +28,7 @@ meta_descr: meta_descr.roadmap
                 <div class="tabPanel-widget">
                     <label for="tab-1" tabindex="0"></label>
                     <input id="tab-1" type="radio" name="tabs" aria-hidden="true">
-                    <h2>2014-2019</h2>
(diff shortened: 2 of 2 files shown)


css/custom.css
```diff
@@ -3805,7 +3805,7 @@ img.emoji {
 
 }
 
-@media only screen and (max-width: 48rem) {
+@media only screen and (max-width: 24rem) {
```
plowsof, line 3808: so after looking into this, what we're doing here is reducing the point where our screens width triggers the "roadmaps mobile view" (which isn't the best) to display by 50% and then we view the original/landscape roadmap, so any visual edits here will carry over to desktop view / vice-versa.

maybe we can move away from max-width and use `@media (orientation: landscape)`
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation

nahuhh: > so after looking into this, what we're doing here is reducing the point where our screens width triggers the "roadmaps ~~mobile~~ inline view" (which isn't the best) to display by 50%* and then we view the original/landscape roadmap,

correct
> so any visual edits here will carry over to desktop view / vice-versa.

not quite

The edits are purely based on the size of the browser window (responsive). On desktop, you can trigger the different levels by resizing the browser window. - these changes don't effect "desktop" or landscape views. It delays the switch of the portrait "tabbed view" to "inline" til a much smaller screen width. 
(the inline view is trash.)

> *im still wondering how the roadmap mobile view styles are applied 

based on the size of the browser window/viewport

> maybe we can move away from max-width and use `@media (orientation: landscape)`
> https://developer.mozilla.org/en-US/docs/Web/CSS/@media/orientation

This doesn't work iiuc, because activating the keyboard will make portrait appear to be landscape

css/custom.css
```diff
@@ -3894,7 +3894,7 @@ img.emoji {
 }
 
 .tabPanel-widget h2 {
-    font-size: 1.4rem;
+    font-size: 1rem;
```
HardenedSteel, line 3897: If you're going to change both values to same 1rem you don't have to include this line (because @ media already won't change anything)

HardenedSteel (changes requested):
If you're going to change both values to same 1rem you don't have to include this line (because @ media already won't change anything)

css/custom.css
HardenedSteel: anyways it can stay at this moment i can clear up this and other codes later

HardenedSteel: #2324 related

HardenedSteel (approved):

plowsof:
we're preventing the broken portrait view from displaying instead of fixing the portrait view and then making one-size-fits-all changes with the abbreviated years, which seems odd, but atm ive not researched how other sites handle this

plowsof:
Ack : to begin a 'responsive' overhaul on the roadmap at this stage in the game with a new site looming is a big ask. this PR's adjustments make the mobile/portrait experience better for the majority of devices out there and removes the redundant "20" in years

plowsof (approved):

PR #2329, opened 2024-07-27
Update qubes whonix wallet isolation user guide

realGWM:


plowsof:
Thanks for this submission, please remove all edits to the none english files.

realGWM:
> Thanks for this submission, please remove all edits to the none english files.

Done.

I guess I misunderstood what "Updates on User Guides" section in the README tells me to do :)

nahuhh:
And please squash, thanks

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
```diff
@@ -1,25 +1,122 @@
...
+
+If you want to update Monero - simply follow all the steps listed in this section again, except the section about configuring the desktop entry and icon.
+
+## 2. Create `monero-wallet-ws` and `monerod-ws` AppVMs:
+
+We need to create two AppVMs to isolate the wallet from the daemon.
+
+* Create `monero-wallet-ws` AppVM that will be used for Monero wallet using whonix-workstation as the template and set networking to none.
+
+* Create `monerod-ws` AppVM that will be used for Monero daemon using whonix-workstation as the template and set networking to sys-whonix. Before moving on, make sure that this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain]({{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
```
realGWM, line 109: The message about checking the size of the raw blockchain is outdated (the file is 86G while actual blockchain is 200G).
It should either be replaced with a hardcoded number (e.g. "make sure that this AppVM has at least 250 gigabytes of private storage") which is not ideal because it will quickly get outdated, or with another source of truth for current blockchain size.

plowsof:
> > Thanks for this submission, please remove all edits to the none english files.
> 
> Done.
> 
> I guess I misunderstood what "Updates on User Guides" section in the README tells me to do :)

No misunderstanding, sorry, its just misleading info @ https://github.com/monero-project/monero-site?tab=readme-ov-file#updates-on-user-guides , that text should be for the maintainer / person who is tasked with handling this. we use tools (po4a combined with weblate) to handle that + it helps with review.

realGWM:
> And please squash, thanks

Done.

Although I will probably need to make a few more commits at least addressing the blockchain size problem I've described above, but I'll try not to forget squashing :)

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
nahuhh: If recommending a pruned node, the blockchain sizendoesnt change as quickly. 

I personally recommend an abundance of free space to account for wear and future.
If someone asks me, i say 128-256gb for pruned node, 512-1tb for full node, and that its better to go big. (write cycles run out slower on a larger drive, and the blockchain grows at rates that cant be predicted)

realGWM: I'm afraid that writing something like "it's recommended to have 512gb-1tb of private storage" might dissuade some/many people from running a full node because this number is too high.

And just to make sure that we are on the same page - in Qubes, "private storage" refers to how much space is allocated to a Qube (a VM), not to the total capacity of your SSD.
Screenshot_2024-07-28_11-59-46
Also due to the way AppVMs work (only /home is persistent), size of files other than the blockchain can be neglected (what I'm trying to say - we don't need to account for "everything else" when writing how much "private storage" a user needs, only blockchain size should be taken into account).
For example, on my system (we ignore everything that is not in /home, as it is not persisted anyway):
```
[workstation user ~]% du -s ~           
209266636	/home/user
[workstation user ~]% du -s ~/.bitmonero
209265992	/home/user/.bitmonero
```
So the total space taken by the Qube is almost the same as size of the blockchain :) 

And also, this "private storage" number can be increased at any time. You can start with 250G, and change it to 300G as soon as you get a notification that your qube has not enough space.

realGWM: I think it might be good idea to mention in the guide that running a pruned node is also an option.

I've never tried that though, is simply adding `--prune-blockchain` to the daemon start params enough?

realGWM: I guess if we don't have a good resource that shows the current size of blockchain (btw I still don't understand why https://moneroj.net/blockchainsize/ shows current blockchain size as ~147G, while my synced blockchain seems to be 200G in size), the best option is to write something like "at the time of writing this guide, the recommended minimum size for a full node is 300G and 100G for a pruned one", which will probably be relevant for a few years? WDYT?

realGWM:
I've added info about pruning, and in absence of better options I replaced advice for estimating amount of space required to download blockchain with hardcoded 300GB recommendation.

This is ready for review now

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
HardenedSteel: > I've never tried that though, is simply adding --prune-blockchain to the daemon start params enough?

It will prune the future blocks, if you want to prune previously downloaded blockchain you have to use pruning tool

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
```diff
@@ -1,25 +1,122 @@
...
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain]({{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
+In this guide we will use the official Monero GUI Qt wallet installed in a whonix-workstation template, but any non-custodial wallet, where you can specify what node to use, should work. You can use a different template for your wallet if you want (and it might even be easier to install if the distribution used for the template contains an up-to-date version of your preferred wallet in the repositories).
 
-## 2. In the AppVM `monerod-ws`:
+We will install both the Monero wallet and daemon in the same whonix-workstation template using pre-compiled binaries. You can simplify the installation process by using the `monero` package available in the Whonix repositories, but it is not up-to-date and also only ships CLI wallet.
```
HardenedSteel, line 17: Is it safe to use distribution packages? See also: https://github.com/feather-wallet/feather/issues/193#issuecomment-2286984927

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
nahuhh: It will prune all* blocks*, but wont shrink the lmbd file size

HardenedSteel: @nahuhh sorry, thanks for correction.

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
nahuhh: Should not use the whonix packaged versions

realGWM: The text here doesn't **endorse** using a packaged version of Monero, merely mentions that it is also an option.  

But let me know if you think it shouldn't be mentioned - I'll remove it.

nahuhh: It should _not_ be used

HardenedSteel (approved):
Looks good to me but I didn't test.

_i18n/en/resources/user-guides/cli_wallet_daemon_isolation_qubes_whonix.md
```diff
@@ -1,25 +1,120 @@
...
+The first step is to install Monero daemon and wallet in the template(s).
 
-  - The first workstation will be used for your wallet, it will referred to as `monero-wallet-ws`. You will have `NetVM` set to `none`.
+Monero daemon needs to be installed in a whonix-workstation template, so that all traffic is routed over Tor.
 
-  - The second workstation will be for the `monerod` daemon, it will be referred to as `monerod-ws`. You will have `NetVM` set to the Whonix gateway `sys-whonix`. Before moving on, make sure this workstation has enough private storage. You can estimate how much space you need by checking the size of the [raw blockchain]({{ site.baseurl }}/downloads/#blockchain). Keep in mind that the blockchain will take up more space with time.
+In this guide we will use the official Monero GUI Qt wallet installed in a whonix-workstation template, but any non-custodial wallet, where you can specify what node to use, should work. You can use a different template for your wallet if you want (and it might even be easier to install if the distribution used for the template contains an up-to-date version of your preferred wallet in the repositories).
```
HardenedSteel, line 15: this section may need to revised

PR #2348, opened 2024-08-15
hangouts: remove facebook link

HardenedSteel:
it wasn't used for years


nahuhh:
The writing style tells me that facebook account is run by the same person who currently posts on twitter.

with 33k followers on facebook, i dont think it should be deleted. Maybe we should _use_ it though

HardenedSteel:
how can we contact?

nahuhh:
I'll try to find out

Issue #2274, opened 2024-03-12
Replacing .svg files


HardenedSteel:
When security level set to safest on Tor browser .svg files don't load. Currently Onion and RSS icons are using SVG file.

plowsof:
thanks didnt know about this, more info here https://tor.stackexchange.com/questions/22229/why-and-how-does-tor-browser-block-svg-assets-when-set-to-safest

jermanuts:
Also Monero Research Lab (MRL) logo is SVG, on https://www.getmonero.org/

https://www.getmonero.org/img/mrl-logo.svg

PR #2359, opened 2024-08-23
svg to png test

plowsof:


HardenedSteel:
closes https://github.com/monero-project/monero-site/issues/2274?

HardenedSteel (commented):
what about other compressed formats? because PNGs will increase the page size a lot

plowsof:
> what about other compressed formats? because PNGs will increase the page size a lot

I made an attempt to compress the png's with some small gains. What other file format would you suggest so i can try to see any gains? we're talking about, in some cases,small 16x16 / 64x64 pixel images so i suspect the gains would be minor. on the other hand, there is the mrl-logo which gets drawn at a whopping 1500x1500 in our css lol not sure what thats about

plowsof:
thanks to @HardenedSteel for exposing me to avif / jpgxl and [srcset](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images)

to attempt to display the not so widely supported jpgxl format and fallback to avif it is as simple as:
```html
    <img class="matrix" 
         srcset="/img/matrix-logo.jxl, /img/matrix-logo.avif" 
         src="/img/matrix-logo.avif" 
         title="Matrix" 
         alt="Matrix logo">
```
all we've done is add the `srcset=` option to the image. We've agreed this would be better handled in a follow up/future PR, that would convert all images to jpgxl and avif + add `srcset` values to all instances of the `` html tag. 

As for this PR i am satisfied with replacing the SVG's with AVIF's with the final TODO being to look at the font files. we can most likely just delete the SVG font files (as the svg is seemingly a fallback for browsers that dont support the standard font types)

PR #2384, opened 2024-10-11
merchants: update p2p/atomic description

nahuhh:


plowsof:
added / adapted in #2426

PR #2405, opened 2024-10-29
update website tech, Astro soon

jermanuts:
Soon! Merge when live website is rebuilt in Astro.

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -412,7 +412,7 @@ contributing:
   gui: "GUI Wallet: "
   gui_p: Mostly QML and C++.
   website: "This website: "
-  website_p: Built with Jekyll. No Javascript.
+  website_p: Built with Astro. No Javascript.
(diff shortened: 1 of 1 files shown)


nahuhh:
NACK

tl strings change with astro

plowsof:
re-open after astro conversion is done. i might even NACK this based on the commit history leading to confusion of who made the conversion.

jermanuts:
I made it a draft, won't probably cause a confusion as the PR for Astro conversion will be probably called something like "convert website to Astro"

plowsof:
The structure of files (especially translation) is going to change. 

It's possible there wont even be file named `_i18n/en.yml`. keeping this PR open, even as a draft, and showing any approval could cause the team working on the new astro site emotional distress thinking they are required to create a merge request that blends into current master branch. 

Add your changes to the astro site / branch where development is taking place. not here. 

There may also be an onus on the current site maintainer to adapt all PR's made to the "future astro" site and handle merge conflicts. For this particular PR it would all be because of a 1 word patch.

PR #2259, opened 2024-03-08
workgroups: remove #monero-dev from Core Team

HardenedSteel:
#monero-dev were listed in Core Team section while its not the core chatroom.

PR #2422, opened 2024-11-15
replace monero-pools with -mining + localised

plowsof:


HardenedSteel (approved):

HardenedSteel:
will cause conflict for #2259

PR #2424, opened 2024-11-20
downloads: link to hangouts for help/support

HardenedSteel:
channels such as `#monero-dev` and `#monero-gui` aren't for user support.

linked to hangouts page

### DIFF
--- a/_i18n/en.yml
+++ b/_i18n/en.yml
@@ -258,8 +258,7 @@ downloads:
   helpsupport: Help and Support
   helpsupport1: "A guide with an explanation of every section of the wallet is available:"
   helpsupport2: "See latest release"
-  gui_helpsupport: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can find the GUI team at #monero-gui, or else check out the Hangouts page for a more complete list of contacts and chatrooms"
-  cli_helpsupport: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can find the CLI team at #monero or #monero-dev, or else check out the Hangouts page for a more complete list of contacts and chatrooms"
+  helpsupport3: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can join to <code>#monero-support</code> or check out the <a href=\"/community/hangouts/\">Hangouts</a> page for a more complete and detailed list of contacts and chatrooms"
   localremote: Local or remote node
   localremote1: Use your own copy of the blockchain or a publicly available one
   transacttor: Transactions over Tor/I2P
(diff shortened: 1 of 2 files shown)


nahuhh:
#monero-support is for support

HardenedSteel:
`#monero-support` added with #2259

nahuhh:
Yes but gui and cli help should be directed to #support, not /hangouts

_i18n/en.yml
```diff
@@ -258,8 +258,7 @@ downloads:
   helpsupport: Help and Support
   helpsupport1: "A guide with an explanation of every section of the wallet is available:"
   helpsupport2: "See latest release"
-  gui_helpsupport: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can find the GUI team at #monero-gui, or else check out the Hangouts page for a more complete list of contacts and chatrooms"
-  cli_helpsupport: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can find the CLI team at #monero or #monero-dev, or else check out the Hangouts page for a more complete list of contacts and chatrooms"
+  helpsupport3: 'If you are experiencing issues or you need more info, feel free to reach out to the community. You can join to <code>#monero-support</code> or check out the <a href="/community/hangouts/">Hangouts</a> page for a more complete and detailed list of contacts and chatrooms'
```
nahuhh, line 261: Most of the other strings don't use quotes, and the ones do, seem to prefer double quotes. Probably best to stick with existing code style


also, syntheticbird had mentioned that we should avoid putting HTML into the strings

HardenedSteel: how can we hyperlink other than using HTML code?

nahuhh: I asked and didnt get a response. I dont know 😅

plowsof: we can use our own jekyll-multiple-languages-plugin, with a [small modification](https://github.com/monero-project/monero-site/commit/6599e723d5fafde628a1c2a3f97cff05c0d3f41b) to enable `{{link:global_url:translated_title}}`, this is also useful when a website goes offline and we can just change it in one place and have it propagate throughout the translations. the path for that is #2454 -> #2237 -> add the above modification -> go wild finding all href links. just not tested with po/po4a.

The reason why this has single quotes is because of the href containing double quotes. these have to be `\"` escaped to prevent crash during build.
  `helpsupport3: "[...] <code>#monero-support</code> or check out the <a href=\"/community/hangouts/\">Hangouts</a> [...]"`

As for the inline `<code></code>` D:   ... so, po4a digests markdown (which could contain `code` stuff) and has a "The string below is a code block" (without these tags) then spits out the html later with the tags. po4a does not touch these .yml files here so it's 'needed' i guess. and might help explain why html may end up being interpreted as plain text depending where it is. i hope i've some sense

plowsof, line 261:
```suggestion
  helpsupport3: "If you are experiencing issues or you need more info, feel free to reach out to the community. You can join to <code>#monero-support</code> or check out the <a href=\"/community/hangouts/\">Hangouts</a> page for a more complete and detailed list of contacts and chatrooms"
```

plowsof:
please squash

plowsof (approved):

PR #2449, opened 2025-03-15
po4a: translation updates

plowsof:
yolo


plowsof:
used as a citation from moving away from po/pot files on repo

PR #2454, opened 2025-03-22
deployment/README: Dockerise

plowsof:


plowsof:
ready for review / testing  :pray:

Gemfile.lock
```diff
@@ -1,32 +1,37 @@
...
       addressable (~> 2.4)
+      base64 (~> 0.2)
       colorator (~> 1.0)
+      csv (~> 3.0)
       em-websocket (~> 0.5)
       i18n (~> 1.0)
       jekyll-sass-converter (>= 2.0, < 4.0)
       jekyll-watch (~> 2.0)
+      json (~> 2.6)
       kramdown (~> 2.3, >= 2.3.1)
```
nahuhh, line 31: can ones like this be simplified to
```suggestion
      kramdown (~> 2.3.1)
```

Gemfile.lock
```diff
@@ -1,32 +1,37 @@
...
       kramdown-parser-gfm (~> 1.0)
       liquid (~> 4.0)
-      mercenary (>= 0.3.6, < 0.5)
+      mercenary (~> 0.3, >= 0.3.6)
```
nahuhh, line 34:
```suggestion
      mercenary (~> 0.3.6)
```

Gemfile.lock
```diff
@@ -41,32 +46,33 @@ GEM
...
+    kramdown (2.5.1)
+      rexml (>= 3.3.9)
     kramdown-parser-gfm (1.1.0)
       kramdown (~> 2.0)
-    liquid (4.0.3)
-    listen (3.7.1)
+    liquid (4.0.4)
+    listen (3.9.0)
       rb-fsevent (~> 0.10, >= 0.10.3)
       rb-inotify (~> 0.9, >= 0.9.10)
```
nahuhh, line 57:
```suggestion
      rb-fsevent (~> 0.10.3)
      rb-inotify (~> 0.9.10)
```

Gemfile.lock
plowsof: I have no idea, tbh i don't even know what these symbols mean, i will read this and have a think https://blog.saeloun.com/2022/08/16/understanding-gemfile-and-gemfile-lock/#what-are-the-various-configuration-options-for-a-gem

nahuhh: AIUI

`~>` means use this major version
so `~> 2.1` means use 2.1+ and less than 3.0

nahuhh: https://blog.saeloun.com/2022/08/16/understanding-gemfile-and-gemfile-lock/#pessimistically-greater-than-or-equal-to-symbols

plowsof: yep, i see they are functionally the same, good to know. 

these are generated automatically though by `bundle update`. I just tried this change, then re-ran the update - sadly it ignores the change and re-adds the previous, so this would just make gem updates tedious

plowsof: I see i'm not locking the bundler version the Dockerfile uses, i've fixed this with `RUN gem install bundler:2.2.23`  so thanks for that - deleting the entire lock file and regenerating it gives the same result, so we don't actually know what version of ruby/bundler created this lockfile previously, another benefit of Docker

plowsof:
when we build the container it mounts eeeeeeverything into the build context, `.dockerignore` prevents this , the files in there are not required to `build`

catduck101 (approved):

PR #2461, opened 2025-04-02
Downloads: add trezor safe 5 hww

### LINKED ISSUE #2456: Please add Trezor Safe 5 in " https://www.getmonero.org/downloads/ " page.  ( Monero software has already supported for Trezor Safe 5. )
Please add Trezor Safe 5 in " https://www.getmonero.org/downloads/ " page. 
( Monero software has already supported for Trezor Safe 5. )

Image

plowsof:
closes #2456


nahuhh (changes requested):
Line 393

plowsof:
Ah thanks, mobile-only view strikes again......  lol we're missing safe 3 there too. will fix here

nahuhh (approved):

PR #2477, opened 2025-04-30
tools: remove monerovision.com / exploremonero, add monerowat.ch

plowsof:


nahuhh:
Has been 7yrs since repo updated. Maintainer appears to be inactive

can you hit the button to rerun tests

plowsof:
Exploremonero apparently do not host their own node and use xmrchain.net api. If true id delist them. Blockexplorer means you are hosting a node

plowsof:
https://xmrchain.net/ "For a pretty version of this explorer with javascript. NOTE - this doesn't work when I turn the API off. " And its not operational currently. I personally switched the twatter general.fund alert to point to p2pools.explorer cus it was broken many.months ago instead of explore monero

nahuhh:
I think exploremonero needs to host there own node

theres also monerowat.ch (seems to be broken atm)

nahuhh (approved):

PR #2504, opened 2025-07-25
Improve italian translation

everoddandeven:


nahuhh:
Can you compare to en.yml and add missing keys, remove keys not present in en.yml, and sort? 

Weblate is supposed to do this (keeps keys in sync) but hey, weblate is offline 🙃


pinging @plowsof for ack/nack on the above

nahuhh (commented):
very fast scan for nits. Will take a better look later

nahuhh:
Please put english changes in a separate pr

PR #2548, opened 2025-10-27
add haveno and retoswap on merchants & exchanges page

fxrstor:
As the title suggests, this PR adds Haveno and Retoswap on merchants & exchanges page.
It also cleans up redundant localmonero references in i18n translation files.

All the translations are machine-translations and are translated using DeepL (EN, DE, PL, IT are human verified).


nahuhh:
@fxrstor, i think the haveno logo is too small.

i didnt download it to view the resolution

PR #2550, opened 2025-10-29
merchants: list retoswap

### LINKED ISSUE #2304: Add Haveno.exchange in P2P exchanges and Atomic Swaps and remove the localmonero
In the menu under the > Community > Merchants & Exchanges > P2P exchanges and Atomic Swaps to add Haveno.exchange include the link to there website as it now the only nonkyc fiat to xmr on and off ramp. Then subheading inside the Haveno.exchange a list of the haveno networks which is currently live and working network:
Haveno-Reto network and a link to there github page:

https://github.com/retoaccess1/haveno-reto

Also need to remove localmonero as has shutdown

fxrstor:
Should fix #2304


fxrstor:
@HardenedSteel Would you approve of this?

HardenedSteel (commented):
Better to keep only human reviewed translations.

For approving Retoswap better to make do some research customer satisfaction, operators' professionalism, knowledge, dispute handling etc.

AFAIK Retoswap team lost their access to their group chat in the past. Handles disputes late.

preland (commented):
I can’t confirm that the translations are accurate, but I don’t see any glaring issues with this at a cursory look

fxrstor:
@plowsof Need your 2c.

community/merchants/index.md
```diff
@@ -86,6 +86,25 @@ meta_descr: merchants.descr
           </div>
         </div>
       </div>
+      <div class="right half no-pad-sm col-lg-6 col-md-6 col-sm-12 col-xs-12">
+        <div class="info-block">
+          <div class="center-xs">
+            <img class="merch" src="/img/retoswap.png" alt="RetoSwap Logo" title="RetoSwap">
```
nahuhh:
```suggestion
            <img class="merch" src="/img/merchants/retoswap.png" alt="RetoSwap Logo" title="RetoSwap">
```

and also move the file

nahuhh:
at this point in time, i'm neutral

plowsof (approved):

HardenedSteel:
Rules aren't same as process of a trade or protocol, check the Bisq's for an example: https://bisq.wiki/Table_of_penalties

preland:
Haveno’s rules *are* the protocol. When a transaction is started through Haveno, both parties agree to the terms of the transaction. If both parties are satisfied with the outcome of the transaction, then it is closed. If a party is not satisfied, then an arbitrator can be allowed to determine the outcome.

If you are talking about more general/conduct rules, Haveno doesn’t have those.

And considering that Erciccione was one of the main early contributors to Haveno…..had it attempted to do so, it likely would’ve been thrown out around the time that they left the project.

HardenedSteel:
> Haveno’s rules are the protocol.
Yeah but we're talking about Retoswap not Haveno protocol.

If you meant Retoswap's rules are Haveno's protocol. The protocol doesn't explain how fiat payments have to be. 

---

> general/conduct rules, ~~Haveno~~ Retoswap doesn’t have those.

It has to, at least in basic level. Examples can be found on Bisq's page.

Also Retoswap has non-written rule third party payments are not allowed (like any other p2p market including Bisq and Localmonero), so Retoswap has rules and they're **NOT** Haveno's protocol.

---

> If a party is not satisfied, then an arbitrator can be allowed to determine the outcome.

So it depends on arbitrator's mood.

preland:
I’ve been intentionally focusing the scope of my discussion towards that of Haveno instead of Retoswap, as I do not know if they have implemented any additional checks or rules since their initial launch.

I would go into more detail about my many gripes about Reto, but I’ve been yelled at in the past by other community members because of speaking against it, so I’ll just stop myself at saying that I would have some personal concerns if I had to actively use it 🤷‍♂️


If anyone actually affiliated with RetoSwap could chime in on this, that would be highly appreciated, as I don’t personally feel like trying to figure out what’s going on over there atm😊

_i18n/es.yml
```diff
@@ -198,6 +200,9 @@ merchants:
   hodlhodldescr: Plataforma de intercambio Peer-to-Peer con fideicomiso sin custodia.
   locmondescr: Plataforma de intercambios Peer to Peer en línea que ofrece intercambios
     de persona a persona.
+  retoswapscr: Una instancia pública del protocolo de intercambio P2P descentralizado Haveno,
```
catduck101: I think you missed removing this one?

ghost (approved):

PyXMR2025 (approved):

plowsof:
i prefer #2592

PR #2594, opened 2026-02-16
Add XMR to USDC to swappers list

xmrtousdc:
This PR adds xmrtousdc.xyz to the list of cryptocurrency swappers on the Merchants & Services page.

**Service Details:**
- Name: XMR to USDC
- URL: https://xmrtousdc.xyz/
- Service Type: Non-custodial instant exchange
- Supported Pairs: XMR ↔ USDC
- No KYC required
- Fast, automated swaps

This service provides a privacy-focused way for users to exchange Monero for USDC and vice versa.

### DIFF
--- a/community/merchants/index.md
+++ b/community/merchants/index.md
@@ -108,6 +108,7 @@ meta_descr: merchants.descr
             <li><a href="https://changenow.io/">ChangeNow</a></li>
             <li><a href="https://godex.io/">Godex</a></li>
             <li><a href="https://stealthex.io/">StealthEX</a></li>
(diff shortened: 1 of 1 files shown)


mosin311aa:
Question as you seem to be the person running this, why exactly should this be used, when the average user can also just use Wagyu directly instead of relying on your wrapper? Using this just introduces an additional middleman in the transaction, which conflicts with basic privacy etiquette.

I don't mean to sound rude, but it seems like you just threw together a page using AI and slapped the Wagyu API (as stated on your website) on it.

plowsof:
Unsurprisingly the average user isn't typing wabooboo into a search engine so you get these grifters with SEO friendly urls.

PR #2606, opened 2026-03-13
ci: bump checkout action to v6

plowsof:
validate hashes has v2 which has some scary warnings about deprecation so bumped em all while we're here
https://github.com/actions/checkout?tab=readme-ov-file#checkout-v6


plowsof:
Thanks for the feedback! :+1:

nahuhh (approved):

PyXMR2025 (approved):

PR #2613, opened 2026-03-23
ci: optimize CI, create source preparation script [beta]

### LINKED ISSUE #2610: [beta] ci build optimization
The ci builds the website in 7mins, includes all blog posts etc. This should be possible to reduce by 60+%.

Should we have it build using docker, and reduce the number of blog posts?

also, is there any reason its uploading build artifacts?

redsh4de:
Closes #2610 

### Changes
Seperate Build / Validate jobs into a single Build & Validate job, skipping upload step

Run build / lint in parallel to each other

Adds a prepare script that CI uses to do the following:
- Disable OpenGraph thumbnail generation
- Keep up to N recent blog posts (including any edited ones from the past), set to 10 currently
- Only render default + edited locales

CI also additionaly disables image optimization seperately

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,9 +50,8 @@ jobs:
         run: pnpm check
 
   build:
-    name: Build
+    name: Build & Validate
     runs-on: ubuntu-22.04
-    needs: [lint]
 
     steps:
       - uses: actions/checkout@v6
@@ -65,45 +64,24 @@ jobs:
         with:
           node-version: "22"
           cache: "pnpm"
-
-      - name: Install production dependencies
-        run: pnpm install --frozen-lockfile --prod
-
-      - name: Build site
-        run: pnpm build
-
-      - name: Upload build artifacts
-        uses: actions/upload-artifact@v7
-        with:
-          name: site-build
-          path: dist/
-          retention-days: 1
-
-  validate:
-    name: Validate HTML
-    runs-on: ubuntu-22.04
-    needs: [build]
-
-    steps:
-      - uses: actions/checkout@v6
-
-      - name: Setup pnpm
-        uses: pnpm/action-setup@v4
-
-      - name: Setup Node.js
-        uses: actions/setup-node@v6
-        with:
-          node-version: "22"
-          cache: "pnpm"
-
-      - name: Download build artifacts
-        uses: actions/download-artifact@v8
-        with:
-          name: site-build
-          path: dist/
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
+
+      - name: Prepare source for CI build
+        env:
+          BASE_BRANCH: ${{ github.event.pull_request.base.ref }} # PR target branch
+          BASE_REF: ${{ github.event_name == 'push' && github.event.before || '' }} # previous commit before the push event
+        run: |
+          ARGS="--limit-posts 10 --skip-og --limit-locales"
+          if [ -n "$BASE_BRANCH" ]; then ARGS="$ARGS --base-branch $BASE_BRANCH"; fi
+          if [ -n "$BASE_REF" ]; then ARGS="$ARGS --base-ref $BASE_REF"; fi
+          pnpm prepare-build $ARGS
+
+      - name: Build site
+        env:
+          SKIP_IMAGE_OPTIMIZATION: "true"
+        run: pnpm build
 
       - name: Validate built HTML
         run: pnpm exec htmlhint 'dist/**/*.html'
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ All commands are run from the root of the project:
 | :--------------------- | :----------------------------------------------- |
 | `pnpm install`         | Installs dependencies                            |
 | `pnpm dev`             | Starts local dev server at `localhost:4321`      |
+| `pnpm prepare-build`   | Trim sources before `pnpm build` (see below)     |
 | `pnpm build`           | Build your production site to `./dist/`          |
 | `pnpm preview`         | Preview your build locally, before deploying     |
 | `pnpm lint`            | Run ESLint to check for code issues              |
@@ -50,6 +51,31 @@ All commands are run from the root of the project:
 | `pnpm format`          | Format code using Prettier                       |
 | `pnpm astro ...`       | Run CLI commands like `astro add`, `astro check` |
 | `pnpm astro -- --help` | Get help using the Astro CLI                     |
+
+## Prepare build
+
+It is possible to trim the source code before `pnpm build` to speed up preview deployments (fewer blog posts, fewer locales, no OG image generation).
+
+```bash
+pnpm prepare-build [flags]
+```
+
+Each flag has an env var equivalent, for build environments that don't accept CLI arguments.
+
+| Flag              | Env var              | Description                                               |
+| :---------------- | :------------------- | :-------------------------------------------------------- |
+| `--limit-posts N` | `LIMIT_POSTS=N`      | Keep only the N most recent blog posts                    |
+| `--skip-og`       | `SKIP_OG=true`       | Remove OpenGraph image generation                         |
+| `--limit-locales` | `LIMIT_LOCALES=true` | Build only the default locale (+ locales with PR changes) |
+
+Changed-file detection (pick one):
+
+| Flag                 | Env var            | Description                                              |
+| :------------------- | :----------------- | :------------------------------------------------------- |
+| `--base-branch NAME` | `BASE_BRANCH=NAME` | Branch to diff against via merge-base (for PRs/previews) |
+| `--base-ref SHA`     | `BASE_REF=SHA`     | SHA to diff against directly (for push events)           |
+
+`SKIP_PREPARE_BUILD=true` early-exits the script. Useful if you can't set a per-environment build command in your platform, and have to skip the script in production.
 
 ## Docker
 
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -66,44 +65,21 @@ jobs:
...
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Prepare source for CI build
+        run: >-
+          pnpm prepare-ci
+          --limit-posts 10
+          --skip-og
+          --limit-locales
+          --base-sha ${{ github.event.pull_request.base.sha }}
```
plowsof: does not work when ci is ran from a push

redsh4de: added a fallback for push events

.github/workflows/ci.yml
```diff
@@ -50,9 +50,8 @@ jobs:
         run: pnpm check
 
   build:
-    name: Build
+    name: Build & Validate
     runs-on: ubuntu-22.04
-    needs: [lint]
```
nahuhh, line 55: Imo should remain

README.md
```diff
@@ -112,11 +112,20 @@ Then open [http://127.0.0.1:4321](http://127.0.0.1:4321).
 
 | Arg                        | Description                                         | Example |
 | :------------------------- | :-------------------------------------------------- | :------ |
-| `LIMIT_POSTS`              | Keep only the N most recent blog posts for previews | `6`     |
+| `PREPARE_CI_ARGS`          | Arguments for the CI preparation script             | `--limit-posts 6 --skip-og` |
```
nahuhh: Since this isnt only for CI, would something like `BUILD_FLAGS` be more accurate?

plowsof:
confirmed "Keep up to N recent blog posts (including any edited ones from the past), set to 10 currently" did not sanity check the edits to translated pages yet

.github/workflows/ci.yml
redsh4de: neutral on this - it's not a significant improvement, shaves down just about a minute since it lets us run lint and build in parallel

README.md
redsh4de: Yeah, good call - the only consumer is the prepare-ci script but should make it more neutral

redsh4de: PREPARE_BUILD_ARGS works better for docker

.github/workflows/ci.yml
plowsof: wanting to test this in coolify. because it only? has access to the [source_commit](https://coolify.io/docs/knowledge-base/environment-variables#source-commit) previews would only see the changed files in the latest commit, and not render any other changes for previews. 

all PR's would have to be squashed into 1 commit to be sure the preview displays correctly if they are touching multiple localised files blogs outside of the 10 latest :thinking: 

i note that existing site doesn't support previews for blogs older than the 10 latest. 

does vercel have access to githubs ci variables? free? limitations? https://github.com/redsh4de/monero-site/deployments

redsh4de: I see, for Coolify in this case would be safer to diff against the whole branch instead of the latest sha?.... git merge-base agains the target should prevent us needing to squash the commits

Vercel doesn't take github's CI into account, needed to set up a custom build command for previews that runs the script - not sure if Coolify will require the same

nahuhh: I think the status-quo for coolify is likely fine (just limit to latest 10 blog posts) and can probably omit --limit-locales (from coolify)

nahuhh:
for coolify using docker build

| time | limit-posts | limit-locales | skip-og | image optimization |
| ---- | ---- | ---- | ---- | ---- |
| 2:24 | 10 | yes | yes | skip |
| 2:19 | 10 | no | yes | skip |
| 2:29 | 10 | no | no | skip |
| 2:33 | 100 | no | no | skip |
| 4:46 | no | no | no | no |

.github/workflows/ci.yml
```diff
@@ -66,44 +65,21 @@ jobs:
...
+        run: >-
+          pnpm prepare-ci
+          --limit-posts 10
+          --skip-og
+          --limit-locales
+          --base-sha "${{ github.event.pull_request.base.sha || github.event.before }}"
 
       - name: Build site
+        env:
+          SKIP_IMAGE_OPTIMIZATION: "true"
```
plowsof, line 83: coolify has separate ENV variables for production and preview deploys, but seemingly only 1 Build command for all. 

so if these options where all ENV variables (or either/or - e.g. the --flags override the env variables) it would make things convenient, unless im missing something

scripts/prepare-build.ts
```diff
@@ -0,0 +1,146 @@
...
+
+if (args["base-branch"] && args["base-ref"]) {
+  console.error("Cannot use both --base-branch and --base-ref");
+  process.exit(1);
+}
+
+const BLOG_DIR = "src/content/blog";
+const OG_ROUTE = "src/pages/open-graph/[...route].ts";
+const I18N_CONFIG = "src/i18n/config.ts";
+const I18N_DIR = "src/i18n/translations";
```
plowsof, line 37: afaict we still need to rebuild when moneropedia entries are edited? i have made a patch and only confirmed via the log line output from the ci run of prepare script. i will add it as a separate commit at the end of this or other PR when ready https://github.com/plowsof/monero-site/commit/65f8289ab1b79282d0786df5a1bbe824ee243540

plowsof:
> for coolify using docker build
> 
> time	limit-posts	limit-locales	skip-og	image optimization
> 2:24	10	yes	yes	skip
> 2:19	10	no	yes	skip
> 2:29	10	no	no	skip
> 2:33	100	no	no	skip
> 4:46	no	no	no	no

i would agree the status quo is fine for coolify, can probably ignore limit-locales here, unless the test commit had touched many translation files, skip-og/images is great.

.github/workflows/ci.yml
redsh4de: added a `--base-branch` flag that can be used with coolify, instead of a commit hash

.github/workflows/ci.yml
redsh4de: hm thats unfortunate... in this case have to pass envs for all

nahuhh: I think 
- pr previews = auto via coolify
- site deploy = manual via core

for time being, can "clone" the coolify project and use one for pr previews and the other for the actual deployment

plowsof: yes site deploy by core **should** be Docker -> generates raw html files -> upload + extract the zip file. should be treated as us trying to gain access to general fund vaults and critical infra, precautions should be taken. 

it will still be useful to have a deploy using coolify to make sure all build routes are run when merges lag behind deploy, i will keep monerodevs running like that.

redsh4de: added env var support, can now set them per-environment

build command can be something like `pnpm prepare-build && pnpm build`

README.md
```diff
@@ -112,11 +112,26 @@ Then open [http://127.0.0.1:4321](http://127.0.0.1:4321).
 
 | Arg                        | Description                                         | Example |
 | :------------------------- | :-------------------------------------------------- | :------ |
-| `LIMIT_POSTS`              | Keep only the N most recent blog posts for previews | `6`     |
+| `PREPARE_BUILD_ARGS`       | Arguments for the source preparation script         | `--limit-posts 6 --skip-og` |
 | `SKIP_IMAGE_OPTIMIZATION`  | Disable image processing for faster builds          | `true`  |
 
+`PREPARE_BUILD_ARGS` are passed to `scripts/prepare-ci-build.ts`:
```
plowsof:
```suggestion
`PREPARE_BUILD_ARGS` are passed to `scripts/prepare-build.ts`:
```

plowsof:
ENV vars im using to test this:

- build command `pnpm build`
- env variables for deploy:
    - none other than NIXPACKS_NODE_VERSION `22`
- env variables for preview
    - PREPARE_BUILD_ARGS `--limit-posts 10 --skip-og`
    - SKIP_IMAGE_OPTIMIZATION `true`

plowsof:
correct usage in coolify:
- build command `pnpm prepare-build && pnpm build`
- env variables for deploy:
    - none other than NIXPACKS_NODE_VERSION 22
- env variables for preview
    - LIMIT_POSTS=10
    - NIXPACKS_NODE_VERSION=22
    - SKIP_IMAGE_OPTIMIZATION=true
    - SKIP_OG=true

nahuhh:
```
PREPARE_BUILD_ARGS --limit-posts 10 --skip-og
SKIP_IMAGE_OPTIMIZATION true
```

this works for the dockerfile deployment

```
env variables for preview
LIMIT_POSTS=10
NIXPACKS_NODE_VERSION=22
SKIP_IMAGE_OPTIMIZATION=true
SKIP_OG=true
```

are you sure this isnt building an older version of this pr? I thought `LIMIT_POSTS` env var was removed. (maybe just for docketfile though

mine looks like:

scripts/prepare-build.ts
```diff
@@ -0,0 +1,174 @@
...
+  options: {
+    "limit-posts": { type: "string" },
+    "skip-og": { type: "boolean", default: false },
+    "limit-locales": { type: "boolean", default: false },
+    "base-branch": { type: "string" },
+    "base-ref": { type: "string" },
+  },
+});
+
+const limitPosts = args["limit-posts"] ?? process.env.LIMIT_POSTS;
```
plowsof, line 22: LIMIT_POSTS easter egg is here

redsh4de:
> are you sure this isnt building an older version of this pr? I thought `LIMIT_POSTS` env var was removed. (maybe just for docketfile though

yeah, PREPARE_BUILD_ARGS is meant for Docker-only - docker passes them as args to the script

you'd use the LIMIT_POSTS env (and other envs) when not using docker, meant for deployment platforms so they can have per-environment setups

plowsof:
just need to add these ENV vars to the readme also and squashing, i can follow up with the moneropedia patch in another PR. LGTM *thanks for testing the docker build also @nahuhh

redsh4de:
updated the readme w/ the envs, should be more easily readable now

nahuhh (approved):

plowsof (approved):

PR #2633, opened 2026-04-22
sitemap_generator: use File.exist? (removed in Ruby 3.2)

thomasbuilds:
File.exists? has been deprecated since Ruby 2.1 and was removed in Ruby 3.2, causing 'bundle exec jekyll serve' to fail with NoMethodError on any contributor's machine running a current Ruby. The Netlify build is still pinned to Ruby 2.7.1 so production is unaffected, but local development on modern Ruby is broken without this one-character fix.


plowsof:
See #2454 #2237

PR #2635, opened 2026-04-24
Deprecating Monero's Unlock Time feature

j-berman:
I think it would be a good idea for someone with greater familiarity with the site to also link to this post from this page: https://www.getmonero.org/resources/moneropedia/unlocktime.html , and include a red warning on that page that the feature is set to be deprecated.

Also if anyone has any links handy to testimonials from people who lost money because of the feature, that would be appreciated in place of [this link](https://github.com/monero-project/research-lab/issues/78#issuecomment-925312477).

### DIFF
--- a/_posts/2026-04-24-deprecating-unlock-time.md
+++ b/_posts/2026-04-24-deprecating-unlock-time.md
@@ -1,0 +1,23 @@
+---
+layout: post
+title: Deprecating Monero's Custom Transaction Unlock Time
+summary: Announcing the deprecation of Monero's Custom Transaction Unlock Time feature
+tags: [announcements, community]
+author: Justin Berman (j-berman)
+---
+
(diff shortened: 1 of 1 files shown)


plowsof:
we are using a deprecated warning for docs pages. here is what it looks like for unlocktime (title/body/url label) using #2641

j-berman:
Instead of "Something happened", maybe it could explicitly say "Deprecated", but that looks good!

Can we have the link point to a specific blog post e.g. in this case have it point to this blog post, and in other cases point somewhere else?

plowsof:
Yes will do, just example text atm :D yes, bF would then, on the day of deploy, rename the blog post to current date so it can then be linked directly

jeffro256 (approved):

plowsof (approved):

SamsungGalaxyPlayer (approved):

PR #2638, opened 2026-04-24
i18n: update translation links

PyXMR2025:


nahuhh:
Nack. The current domain is temporary

PyXMR2025:
> Nack. The current domain is temporary

We can change it back after https://translate.getmonero.org/ is online.

PR #2640, opened 2026-04-25
ui/components: add PreviewBanner with commit hash for coolify [beta]

plowsof:
coolify sets `SOURCE_COMMIT` env var by default. I have manually set the env `COOLIFY_PREVIEW` to `true` for preview deploys, but not for production

styles can be added e.g. background colour 

the variable can/should be generic?  i seen a nice looking default var called `NODE_ENV` but coolify has this set to `production` in both deploy and previews 

proof of concept to show the commit hash on previews

- linting todo

### DIFF
--- a/src/components/ui/PreviewBanner.astro
+++ b/src/components/ui/PreviewBanner.astro
@@ -1,0 +1,35 @@
+---
+const isCoolifyPreview = import.meta.env.COOLIFY_PREVIEW;
+const sourceCommit = import.meta.env.SOURCE_COMMIT;
+const coolifyBranch = import.meta.env.COOLIFY_BRANCH; // "pull/2639/head:pr-2639-coolify",
+const coolifyRepo = import.meta.env.COOLIFY_REPO;
+if (!isCoolifyPreview) return;
+const coolifyPrNum = coolifyBranch.split("/")[1];
+---
+
+<div class="preview-banner">
+  <span class="preview-banner-style">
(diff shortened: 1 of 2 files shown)


redsh4de:
i like this

this can be made as a component in `components/ui/header/PreviewBanner.astro`, and imported in the main header

then it can own its own functions & styles

plowsof:
i ended up putting it in Layout to try to not have it interfere with the UI with no success. I then realised it should be fixed at the bottom just like how netlify displays something similar without messing with the UI, so im not sure where it should go now. 

coolify has another default var `COOLIFY_BRANCH` which contains the pr number - this allows us to have a direct link to the commit

astro? sometimes {var} sometimes {`${var}`}{var} :shrug: 

TODO: move the banner element inside the html block for linting

src/components/ui/PreviewBanner.astro
```diff
@@ -0,0 +1,35 @@
...
+    text-align: center;
+    z-index: 999;
+  }
+
+  .preview-banner-style {
+    display: inline-block;
+    background: black;
+    color: white;
+    padding: 6px 12px;
+    border-radius: 6px;
```
redsh4de: design tokens from `styles/global.css`:
```suggestion
    padding: var(--space-2xs) var(--space-md);
    border-radius: var(--radius-md);
```

redsh4de (changes requested):

redsh4de (approved):

PR #2646, opened 2026-05-02
components/ui/dropdown/DropDown.astro: add name property to group dropdowns [beta]

plowsof:
implemented on the core downloads page so when cli/gui dropdown is opened, the other will be minimised


src/components/pages/downloads/CoreCard.astro
plowsof, line 124: ive copied the style for `<Button` with things on new lines and ending on `>`

redsh4de (approved):

PR #2653, opened 2026-05-06
ci: confirm dist archive is reproducible [beta]

plowsof:
compare zipped archive hash of dist folder between 2 separate workflow runs upload archive on release
use env variables for consistency
bump actions
only run on beta releases

with this the build runs twice in parallel always. for push/pr optimised ENV variables are set as usual. for release these are not set and the dist folder archive is uploaded there as an artifact.


plowsof:
rebased after #2654 merged

redsh4de (approved):

PR #2654, opened 2026-05-06
remove non determinism from deployed site files [beta]

plowsof:
#2653

3 commits have small diffs, i built with everything, committed the dist folder, removed the commit and rebuilt, the file is the resulting diff

[Accordian.astro-15864bcc.txt](https://github.com/user-attachments/files/27460020/Accordian.astro-15864bcc.txt)
[index.astro-7c1dbd7b.txt](https://github.com/user-attachments/files/27460021/index.astro-7c1dbd7b.txt)
[SectionTabs.astro-457800c3.txt](https://github.com/user-attachments/files/27460022/SectionTabs.astro-457800c3.txt)

### DIFF
--- a/src/components/ui/accordion/Accordion.astro
+++ b/src/components/ui/accordion/Accordion.astro
@@ -1,6 +1,4 @@
 ---
-import "node:crypto";
-
 import AccordionItem from "./AccordionItem.astro";
 
 export interface AccordionItemType {
@@ -15,7 +13,15 @@ export interface Props {
 }
 
 const { items, multiExpand = false, variant = "default" } = Astro.props;
-const accordionId = crypto.randomUUID();
+if (items?.some((i) => typeof i.title !== "string" || !i.title.trim())) {
+  throw new Error("<Accordion> item has an empty title.");
+}
+const accordionId = items?.length
+  ? `${variant}-${[...items.map((i) => i.title).join("")]
+      .slice(0, 7)
+      .map((c) => c.codePointAt(0))
+      .join("-")}`
+  : `${variant}-slot`;
 const accordionName = multiExpand ? undefined : `accordion-${accordionId}`;
 ---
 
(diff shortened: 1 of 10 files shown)


src/components/ui/accordion/Accordion.astro
```diff
@@ -15,7 +13,15 @@ export interface Props {
...
-const accordionId = crypto.randomUUID();
+if (items?.some((i) => typeof i.title !== "string" || !i.title.trim())) {
+  throw new Error("<Accordion> item has an empty title.");
+}
+const accordionId = items?.length
+  ? `${variant}-${[...items.map((i) => i.title).join("")]
+      .slice(0, 7)
+      .map((c) => c.codePointAt(0))
+      .join("-")}`
+  : `${variant}-slot`;
```
redsh4de, line 24: Could add a `stableId` helper, smth like this that works with both strings & lists:
```ts
import { createHash } from "node:crypto";

export const stableId = (input: string | string[], length = 12): string => {
  const data = typeof input === "string" ? input : JSON.stringify(input);
  return createHash("sha256").update(data).digest("hex").slice(0, length);
};
```

then can collapse to smth like this, and apply in NavItem, SectionTabs the same:
```ts
import { stableId } from "@/utils/stableId";
// ...
const titles = items?.map((i) => i.title) ?? [];
if (titles.some((t) => !t.trim())) {
  throw new Error("<Accordion> item has an empty title.");
}
const accordionId = `${variant}-${titles.length ? stableId(titles) : "slot"}`
```

redsh4de (changes requested):

src/components/ui/accordion/Accordion.astro
plowsof: nice refactor , (although adding the import back), could PR it after this as i'd hate to back into diff hell / confirming what works

redsh4de (approved):

PR #2655, opened 2026-05-10
update roadmap

plowsof:


nahuhh:
Was there nothing in '25? Looks odd with it just missing

maybe needs to be added with an "empty" message if nothing was done?

_i18n/en.yml
```diff
@@ -828,7 +828,7 @@ roadmap:
   p2pool: P2Pool released
   haveno: Haveno
 
-  seraphis-jamtis: Seraphis codebase & Jamtis
+  seraphis-jamtis: Seraphis codebase & Jamtis(PQ)
```
nahuhh: Should probably divorce these two, since jamtis isnt related to seraphis and will almost certainly be shipped w/o it

plowsof: OSPEAD? :thinking:

plowsof:
> Was there nothing in '25? Looks odd with it just missing
> 
> maybe needs to be added with an "empty" message if nothing was done?

will add the placeholder for it.. a bunch of fcmp++ related audits if note worthy

_i18n/en.yml
plowsof: the reason this was originally added is that fcmp++ was built upon the "seraphis codebase" @ https://github.com/seraphis-migration/monero , not the 'seraphis' with 200+ decoys etc 🤔 confusing

nahuhh: Technically fcmp++ isnt built on seraphis codebase, its but on a fork of monero-project's master branch. Seraphis work might be on seraphis-migration's master branch (not sure), but fcmp++ branches arent built on that

iirc, seraphis-jamtis was added before fcmp++ and carrot were a thing (and seraphis+jamtis was ghe direction). Now, its fcmp++ & carrot, with a new carrot/fcmp++ compatible jamtis in the works

nahuhh: wrt OSPEAD, isnt that research finished, but not relevant (to implement) in the way initially intended since fcmp++ gets rid of some of the issues

rbrunner7: The FCMP++ plus Carrot codebase has indeed nothing to do with the Seraphis + Jamtis codebase and is not based on that. It was just convenient to recycle an already existing GitHub org as a home for the projects, that's why "seraphis" still occurs in the URLs.

I would say that as of today it's completely unclear whether we will ever use parts of the (excellent) Seraphis codebase for anything. Maybe some parts of it for a totally new successor to `wallet2`, but who knows.

Tevador took Jamtis as a base and is developing it further into something that could provide some protection against working quantum computers. I think working title is indeed *Jamtis PQ*. More info here: https://github.com/monero-project/research-lab/issues/151

I propose to give Jamtis PQ a mention in the roadmap, and drop Seraphis completely.

PR #2666, opened 2026-05-20
moneropedia: append .html to unlocktime deprecated url

plowsof:
*still an issue of .html being stripped. i will circle back to this

it worked fine locally so i assumed it was a problem on netlifys end


plowsof:
i sanity checked this and the .html remains locally - i just didnt have the PR pulled ;_;

nahuhh (approved):

jpk68 (approved):

PR #2675, opened 2026-06-26
styles: fix external link wrap [beta]

redsh4de:
makes it so that when the external link icon wraps, it takes the last word of the link with it


nahuhh:
commit title prefix `styles:` please and thanks

redsh4de:
updated

nahuhh (approved):

PR #2676, opened 2026-06-26
roadmap: add Bulletproofs* to roadmap [beta]

redsh4de:


src/i18n/translations/en/roadmap.json
```diff
@@ -30,6 +30,10 @@
             "title": "Jamtis",
             "description": "Post-quantum addressing protocol for Monero."
         },
+        "bulletproofsStar": {
+            "title": "Bulletproofs*",
+            "description": "Folding Bulletproofs for lighter many-input transactions."
```
nahuhh, line 35: I'm not sure if this is the correct way to describe, but otherwise LGTM

nahuhh (approved):

nahuhh:
commit title prefix `roadmap:` plz and thx

src/i18n/translations/en/roadmap.json
redsh4de: What about "A folding scheme for more compact FCMP++ transactions"?

It's not the only thing it enables (theres cheaper tx uniformity as well), but to higlight the biggest impact

plowsof (approved):

PR #2688, opened 2026-07-24
index: center the mark behind the hero

thomasbuilds:
Centers the M behind the hero on the window, and gives the hero the whole first screen so the next section is no longer visible before you scroll.

follow-ups:
* Anchor targets each guess how much of the viewport the header covers, and disagree: 6.5rem, 4rem, 4em, 3rem, 2rem and unset across five files. At 414px `#about-monero` and `#get-involved` have no offset at all, so the header hides the top 40px of their headings.
* `main` applies `--page-top-margin` as both a margin and a padding, so it carries 64px of top spacing rather than 32px. That is why the hero's negative margin here subtracts it twice.
* `--header-height` stays a declared constant, since CSS cannot measure the header. `block-size` at least makes it binding rather than a floor.

### DIFF
--- a/src/components/pages/index/Hero.astro
+++ b/src/components/pages/index/Hero.astro
@@ -109,7 +109,9 @@ const { t, localizeHref } = Astro.locals;
 
   .background-logo {
     position: absolute;
-    inset: 0;
+    inset-block-start: 0;
+    inset-inline: 0;
+    block-size: 100dvh;
     display: grid;
     place-items: center;
     pointer-events: none;
@@ -117,16 +119,9 @@ const { t, localizeHref } = Astro.locals;
   }
 
   .logo-box {
-    inline-size: min(100cqw, 100cqh);
+    inline-size: min(100cqw, 85cqh);
     aspect-ratio: 1 / 1;
     position: relative;
-  }
-
-  @supports not (width: 100cqw) {
-    .logo-box {
-      inline-size: min(90vmin, 100%);
-      aspect-ratio: 1 / 1;
-    }
   }
 
   .mask-icon {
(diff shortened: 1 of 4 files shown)


nahuhh:
Deploy failed cc @plowsof

plowsof:
thanks @nahuhh , the main url was moved to beta.getmonero.org and coolify gave us a helping hand and set the preview url to that without telling me, fixed now

thomasbuilds:
yes nice we can visually compare my change now between this and beta.getmonero.org quickly

nahuhh:
The M behind the hero seems to maintain its position between deploy and germonero.org, but deploy is larger

thomasbuilds:
The M now extends to the available height (this also fixes the incorrect vertical centering). Also on mobile the section wasn't taking up the full height before, so you could see the About section underneath which I think looks ugly. I just re-verified this with iPhone XR in chrome DevTools.

nahuhh (approved):
Approving based on looks

note: not a fan of the verbose "explainer" commit messages

thomasbuilds:
I scoped the PR on what really matters rn and it's easier to review

> What	Link
> 🕵️‍♂️ Review	[Files changed](https://github.com/monero-project/monero-site/pull/2688/files)
> 🌐 Deploy	[Preview #2688](https://2688.beta.monerodevs.org)

too bad this didn't update

jpk68 (approved):

plowsof:
preview is showing the latest now

thomasbuilds:
thanks this new commit also addresses @redsh4de DM
> the overlap of the M logo with the header area was intentional, part of the figma design

redsh4de (approved):

nahuhh (approved):

PR #2707, opened 2026-09-17
pnpm: harden against supply chain attacks

plowsof:
my contribution to security theatre for beta branch / "i looked at https://pnpm.io/supply-chain-security and made a PR"

builds _should fail_ 

- `minimumReleaseAge`: im not proud of this one as hijacked packages can just wait an arbitrary amount of time before revealing themselves but its _something_
- `blockExoticSubdeps`: deps of deps **must come from the npm registry** not from git URLs or raw tarballs that **skip all the other checks**
- `trustPolicy`: a stolen developer login pushes unsigned packages that where previously signed = fail "trust level has decreased compared to previous releases"


plowsof:
one of our deps fails the trustpolicy check , a [workaround](https://github.com/paulmillr/chokidar/issues/1440#issuecomment-4598965491) to exclude the specific package is required, or, update to [v5](https://github.com/paulmillr/chokidar/issues/1440#issuecomment-4601223634)

plowsof:
updated chokidar to v5 . Astro dgaf and just exclude it 

https://github.com/withastro/astro/blob/main/pnpm-workspace.yaml#L70-L71

pnpm-workspace.yaml
```diff
@@ -3,3 +3,8 @@ allowBuilds:
   esbuild: false
   "@parcel/watcher": false
   sharp: false
+minimumReleaseAge: 20160 # minutes (14 days)
```
thomasbuilds, line 6: Suggest 3 days instead of 14.

- In [Woodruff's analysis](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) of 10 major supply-chain attacks, every npm one (chalk, Nx, rspack, web3.js) was pulled within 12h. The longer-lived ones weren't npm (tj-actions, Kong's Docker image, xz), so in that set going past 3d adds nothing pnpm can stop.
- 3d is what MetaMask, Uniswap and Astro use, and what Renovate's `config:best-practices` preset sets for npm (it covers npm's 72h unpublish window).
- pnpm 11 enforces this on every `--frozen-lockfile` install, so any lockfile update that pulls in a release younger than the window keeps CI red until it ages. At 14d CI is red until Sep 24, and security fixes wait two weeks unless excluded.

```suggestion
minimumReleaseAge: 4320 # minutes (3 days)
```

thomasbuilds (changes requested):

SyntheticBird45 (approved):

pnpm-workspace.yaml
plowsof: in security theatre we usually wait at least a week before merging a PR even with approvals. 2 weeks gives us a nice buffer to review every LOC change in each updated package.

thomasbuilds: Imo 2 weeks is excessive, 1 week?

SyntheticBird45: > Imo 2 weeks is excessive, 1 week?

I am against 1 week. 2 weeks is perfectly fine.

redsh4de (approved):

PyXMR2025 (approved):

thomasbuilds:
should be green ci now or tomorrow idk?

plowsof:
thanks for reminder, just checked, in 1 hour~ it should pass, how exciting, ill squash after confirming locally :pray:

